PR #1, opened 2018-05-31
Add support for properly pre-split test and train input data.

Tilps:
Originally: https://github.com/glinscott/leela-chess/pull/688


Tilps:
This should be good to submit, which I'll do tomorrow if someone doesn't review and submit it sooner.

killerducky:
I glanced at this for a few minutes, looks ok to me.

I did notice get_latest_chunks picks based on file timestamp, just something to be aware of.

PR #2, opened 2018-05-31
Change how frequently test batches are run

Tilps:
Originally: https://github.com/glinscott/leela-chess/pull/681


Tilps:
This should be good to submit, which I'll do unless someone else reviews and submits it before tomorrow.

killerducky:
Quickly looked at this, it seems ok.

PR #3, opened 2018-05-31
Normalize 50 move rule plane in training

Tilps:
Originally: https://github.com/glinscott/leela-chess/pull/691

### DIFF
--- a/tf/chunkparser.py
+++ b/tf/chunkparser.py
@@ -66,10 +66,10 @@ class ChunkParser:
         long.
         """
 
-        # Build a series of flat planes with values 0..255
+        # Build 2 flat float32 planes with values 0,1
         self.flat_planes = []
-        for i in range(256):
-            self.flat_planes.append(bytes([i]*64))
+        for i in range(2):
+            self.flat_planes.append(np.zeros(64, dtype=np.float32) + i)
 
         # set the down-sampling rate
         self.sample = sample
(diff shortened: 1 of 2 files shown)


Tilps:
There was one report that this may not be a straight up win - small number of training steps seemed to result in a regression on the win at chess test case.  More testing might be appropriate, or maybe it may just be expected to get a short term regression as some rule 50 overfit gets regularized out.

In any case, review strongly appreciated!

dubslow:
I would be highly suspicious of current network weights being able to adjust to the change immediately. Short term regression isn't cause for concern IMO. What about the various scripts written to adjust weights in sync with this change?

Tilps:
The test case was not with 'current weights' in the way you might expect - it was using net_to_model which with this PR converts the existing (external) weights to work with the reduced input size during training in a way which should mean that same training data will map to same network outputs.  Then after training has run, the export code does the inverse weights transform for producing the new external weights format, so it should work in existing clients.  Based on that 'the regression' 'probably' came from regularization reducing the rule 50 weights faster than training could relearn the things which depended upon their outputs being so large.
There are no external scripts required - weight adjustment is built-in to this PR.

borg323:
A brief history of efforts to address rule50 normalization:
* Originally, there was issue https://github.com/glinscott/leela-chess/issues/602 and @Error323 made https://github.com/glinscott/leela-chess/pull/605 that in addition to the training changes required client support for the normalized plane.
* To alleviate this I made https://github.com/glinscott/leela-chess/pull/655 with two scripts to scale/unscale the coefficients directly in the weights file. I tested this and found that scaled weights file (both ways) with modified client with opposite scaling of the rule50 plane gave identical output to the unscaled weights file with unmodified client. Obviously the output of unmodified client with scaled weights files was different.
* This PR (originally https://github.com/glinscott/leela-chess/pull/691) incorporates the scaling functionality directly into the training code, which I think is a more robust approach.

PR #4, opened 2018-05-31
Port training changes to enable batch normalization betas from leela-zero

Tilps:
Originally: https://github.com/glinscott/leela-chess/pull/690


Tilps:
I've only heard positive things from testers so far for this and its pretty much a straight up port from leela-zero (go) codebase.
Review would be greatly appreciated, but if its not forthcoming in the next couple of days I think I'll be merging this one.

PR #8, opened 2018-06-02
Move process loop into tfprocess.py to better handle resume.

Tilps:
If you resumed from part way through a total_steps count, it would still run for total_steps, which is definitely not ideal.
By moving the process_loop into tfprocess we can easily use the current step count to reduce the number of steps performed so the process will complete on a multiple of total_steps.

### DIFF
--- a/tf/tfprocess.py
+++ b/tf/tfprocess.py
@@ -186,6 +186,14 @@ class TFProcess:
         print("Restoring from {0}".format(file))
         self.saver.restore(self.session, file)
 
+    def process_loop(self, batch_size, test_batches):
+        # Get the initial steps value in case this is a resume from a step count
+        # which is not a multiple of total_steps.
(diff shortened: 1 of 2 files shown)


tf/tfprocess.py
```diff
@@ -186,6 +186,14 @@ def restore(self, file):
         print("Restoring from {0}".format(file))
         self.saver.restore(self.session, file)
 
+    def process_loop(self, batch_size, test_batches):
+        # Get the initial steps value in case this is a resume from a step count
+        # which is not a multiple of total_steps.
+        steps = tf.train.global_step(self.session, self.global_step)
+        total_steps = self.cfg['training']['total_steps']
+        for _ in range(steps % total_steps, total_steps):
```
dubslow: Maybe this is clearer as `range(total_steps - (steps % total_steps))`? (Same result either way)

dubslow:
looked good to me

PR #9, opened 2018-06-02
Fix the output of total loss to represent the actual optimized term.

Tilps:


mooskagh:
What was that about? Looks like adding some random constant without anyone to review/agree.

dubslow:
It was sitting here for several days waiting for anyone at all to review... and no one did

mooskagh:
Catch random passer-bys on #dev and #technical. :)

dubslow:
He tried! twice a day every day.

Tilps:
This was reviewed by KillerDucky - this constant actually makes the massive comment above the line be true - MSE sent to the averager is divided by 4, but the 'loss' which is optimized doesn't have that divide by 4 in place.  This multiplier undoes that so that the total loss actually represents the optimized term.

PR #11, opened 2018-06-05
Update README.md

scs-ben:
added additional options from new config files
added more descriptions on boundaries since it's not clear what they represent

### DIFF
--- a/README.md
+++ b/README.md
@@ -42,7 +42,7 @@ training:
         - 0.02
         - 0.002
         - 0.0005
-    lr_boundaries:                     # list of boundaries (boundaries are absolute to number of trained steps)
+    lr_boundaries:                     # list of boundaries
         - 100000
         - 130000
     policy_loss_weight: 1.0            # weight of policy loss
--- a/README.md
+++ b/README.md
@@ -36,7 +36,6 @@ training:
     batch_size: 2048                   # training batch
     total_steps: 140000                # terminate after these steps
     test_steps: 2000                   # eval test set values after this many steps
-    total_steps: 140000                # terminate after these steps
     # checkpoint_steps: 10000          # optional frequency for checkpointing before finish
     shuffle_size: 524288               # size of the shuffle buffer
     lr_values:                         # list of learning rates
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -26,18 +26,24 @@ gpu: 0                                 # gpu id to process on
...
-  input: '/path/to/chunks/*/draw/'     # supports glob
+  # For separated test and train data.
+  input_train: '/path/to/chunks/*/draw/' # supports glob
+  input_test: '/path/to/chunks/*/draw/'  # supports glob
+  # For a one-shot run with all data in one directory.
+  # input: '/path/to/chunks/*/draw/'
 
 training:
     batch_size: 2048                   # training batch
     total_steps: 140000                # terminate after these steps
```
Tilps: total_steps is listed twice now.

README.md
```diff
@@ -26,18 +26,24 @@ gpu: 0                                 # gpu id to process on
...
+    test_steps: 2000                   # eval test set values after this many steps
+    total_steps: 140000                # terminate after these steps
+    # checkpoint_steps: 10000          # optional frequency for checkpointing before finish
     shuffle_size: 524288               # size of the shuffle buffer
     lr_values:                         # list of learning rates
         - 0.02
         - 0.002
         - 0.0005
-    lr_boundaries:                     # list of boundaries
+    lr_boundaries:                     # list of boundaries (boundaries are absolute to number of trained steps)
```
Tilps: 'Technically' they are absolute modulo total_steps - not sure if we want that detail here or not.  total_steps is also similar, it runs until step count is a multiple of total_steps.
(I don't mind this text as is, just wanted to bring it up in case you have a better idea of what to say as a result.)

README.md
scs-ben: should be removed

README.md
scs-ben: Maybe an explanation somewhere of what the boundaries are in reference to. I want to clarify that they aren't relative to the current run of "total_steps" but absolute to how many steps have been run for the entire process.

scs-ben: something like this under the config example:
* The LR boundaries are in reference to the overall number of steps trained (not relative to the current training run)

Tilps: The part that is potentially confusing is when the current step count is > total_steps.  Which doesn't happen so much in a bootstrap scenario perhaps, but is normal for our ongoing training process.

It can wait till a follow up PR though.

PR #12, opened 2018-06-07
initial new pipeline proposal

Error323:
This update adds the following features:
- utilize ramdisk
- train multiple networks on the same data
- permanent test/train split

It assumes that another script prepares the test/train split and that the underlying `train.py` script utilizes multigpu support in the future. Furthermore it assumes that yet another script will upload the correct network from `$NETDIR` to the server.

Please provide your feedback!

### DIFF
--- a/tf/start.sh
+++ b/tf/start.sh
@@ -5,6 +5,7 @@ set -e
 CONFIGDIR=$1
 GAMES=$2
 
+NETARCHS=(64x6 128x10 192x15)
 ROOT="/work/lc0"
 NETDIR="$ROOT/networks/upload"
 GAMEFILE="$HOME/.lc0.dat"
@@ -23,7 +24,7 @@ file="training.${game_num}.gz"
 echo "Starting with '$file' as last game in window"
 
 train() {
-  unbuffer ./train.py --cfg=$1 --output=$2 2>&1 > "$ROOT/logs/$(date +%Y%m%d-%H%M%S).log"
+  unbuffer ./train.py --cfg=$1 --output=$2 2>&1 | tee "$ROOT/logs/$(date +%Y%m%d-%H%M%S).log"
   gzip -9 $2
   mv -v $2.gz $NETDIR
 }
@@ -36,15 +37,13 @@ do
     echo ""
 
     # prepare ramdisk
-    rm -rf $RAMDISK/*
-    cp -r $ROOT/split/{train,test} $RAMDISK
+    rm -rf $RAMDISK/{train,test}
+    rsync -aq --delete-during $ROOT/split/{train,test} $RAMDISK
 
     # train all networks
-    for netarch in 64x6 128x10
+    for netarch in ${NETARCHS[@]}
     do
-      cfg="$CONFIGDIR/$netarch.yaml"
       echo "Training $netarch:"
-      echo $(cat $cfg)
       train "$CONFIGDIR/$netarch.yaml" "${netarch}-$(date +"%Y_%m%d_%H%M_%S_%3N").txt"
     done
 
@@ -54,7 +53,7 @@ do
     file="training.${game_num}.gz"
     echo "Waiting for '$file'"
   else
-    sleep 60
     echo -n "."
+    sleep 60
   fi
 done
--- a/tf/start.sh
+++ b/tf/start.sh
@@ -2,12 +2,13 @@
 
 set -e
 
-CONFIG=$1
+CONFIGDIR=$1
 GAMES=$2
 
-NET="/tmp/weights.txt"
-NETDIR="/home2/networks/upload"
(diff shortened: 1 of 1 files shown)


tf/start.sh
```diff
@@ -2,12 +2,13 @@
...
 
-CONFIG=$1
+CONFIGDIR=$1
 GAMES=$2
 
-NET="/tmp/weights.txt"
-NETDIR="/home2/networks/upload"
+ROOT="/work/lc0"
+NETDIR="$ROOT/networks/upload"
 GAMEFILE="$HOME/.lc0.dat"
```
mooskagh: If network of one size is trained, but pipeline crashed during training of larger net, restarting script would redo training of smaller network.

I don't see an elegant way to fix that, though, maybe in addition to global GAMEFILE have also per-network-size game network size, and check it in inner loop.

tf/start.sh
```diff
@@ -17,22 +18,43 @@ fi
...
   then
     echo ""
-    unbuffer ./train.py --cfg=$CONFIG --output=$NET 2>&1 | tee /home2/logs/$(date +%Y%m%d-%H%M%S).log
-    TSTAMP=$(date +"%Y_%m%d_%H%M_%S_%3N")
-    FILE="$NETDIR/$TSTAMP.lc0"
-    mv -v $NET $FILE
+
+    # prepare ramdisk
+    rm -rf $RAMDISK/*
+    cp -r $ROOT/split/{train,test} $RAMDISK
```
mooskagh: Will it copy more and more unused games into ramdisk as more data is generated? Maybe there is some way to only copy most recent files according to window size. Or maybe train/test splitting script can do that cleanup and move old games into "archive".

tf/start.sh
```diff
@@ -17,22 +18,43 @@ fi
...
+
+    # train all networks
+    for netarch in 64x6 128x10
```
mooskagh: Maybe make "64x6 128x10" a variable on top of the file?

bash documentation says this is possible:
```bash
farm_hosts=(web03 web04 web05 web06 web07)

for i in ${farm_hosts[@]}; do
```

tf/start.sh
```diff
@@ -17,22 +18,43 @@ fi
 
 game_num=$(cat $GAMEFILE)
 game_num=$((game_num + GAMES))
-echo "Starting with training.$game_num.gz as last game in window"
+file="training.${game_num}.gz"
+
+echo "Starting with '$file' as last game in window"
+
+train() {
+  unbuffer ./train.py --cfg=$1 --output=$2 2>&1 > "$ROOT/logs/$(date +%Y%m%d-%H%M%S).log"
```
Tilps: Using redirect instead of tee means that training is now silent?

tf/start.sh
Tilps: I think the split script should handle this and we should just assume ramdisk is populated.

tf/start.sh
Tilps: I think this is worth fixing - alternative option: a persistent counter which is set to 0 at the end of all training and incremented in the training loop, then check it against a loop counter to decide how many loops can be skipped..

mooskagh: That would break if we want to add/remove network size one day, or change number of games per cycle (very probable)

tf/start.sh
mooskagh: I think we want separate config per network size?

tf/start.sh
Tilps: It should be fine so long as we only modify the script by taking it down when its not training... ?

tf/start.sh
Error323: Indeed. Could change it back though? But I think one should just `tail -f` the log files.

tf/start.sh
Error323: @mooskagh I think the split script should make sure we can do a copy like this.
@Tilps if we don't coordinate here, the data for the various networks will differ, I don't think that's desirable.

tf/start.sh
Error323: @mooskagh that would make everything get out of sync wrt ramdisk data.

Maybe we should in that case remove the last checkpoint for nets that were already trained. At least stay consistent between networks.

tf/start.sh
Tilps: I think you want >> instead of >, if you aren't going to switch back to tee.
I don't have a strong preference, but I think I liked being able to see the progress while the shell script was running without having to work out what todays log date is.

tf/start.sh
Tilps: You make a good point.  I guess rsync would be something to consider.

tf/start.sh
Error323: I'll change back to `tee`

PR #16, opened 2018-06-08
simplify split.sh and respect windowsize

Error323:
Tested on 10K training games.

### DIFF
--- a/scripts/split.sh
+++ b/scripts/split.sh
@@ -1,69 +1,91 @@
-#!/bin/bash
+#!/usr/bin/env bash
 
-NETDIR="/home/folkert/data/run1"
-TSTDIR="/home/folkert/data/work/test"
-TRNDIR="/home/folkert/data/work/train"
+function usage()
+{
+  echo "Uploads a network with NxM prefix, where N=filters and M=blocks"
+  echo ""
+  echo "./split.sh"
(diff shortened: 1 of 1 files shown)


scripts/split.sh
```diff
@@ -1,69 +1,111 @@
...
+      ls -rt $TESTDIR | head -n $overhead_test | xargs -I{} rm -f $TESTDIR/{}
+      ls -rt $TRAINDIR | head -n $overhead_train | xargs -I{} rm -f $TRAINDIR/{}
+      let "n -= $overhead"
+    fi
+  fi
+}
+
+
+echo -n "processing data in '$INPUTDIR'..."
+for f in $(ls $INPUTDIR)
```
Tilps: This needs to be ls -tr I  think.

scripts/split.sh
```diff
@@ -1,69 +1,114 @@
...
+      target=$TESTDIR/$file
+    else
+      target=$TRAINDIR/$file
+    fi
+
+    cp -a $dir/$file $target
+
+    if [ $n -gt $max ]
+    then
+      ls -rt $TESTDIR | head -n $overhead_test | xargs -I{} rm -f $TESTDIR/{}
```
Tilps: should these be 1rt as well?

PR #17, opened 2018-06-10
Add lock, allow split continuing, check file integrity

Error323:
Ok in hindsight I should've split into various commits, but this is what I implemented:
- allow continues when the `$TRAINDIR` and `$TESTDIR` exist
- implement exclusive lock on rsync from `start.sh` and file removal from `split.sh` through `flock`
- implement game file integrity by making sure that file size > 0 and file size is a multiple of 8276

### DIFF
--- a/scripts/split.sh
+++ b/scripts/split.sh
@@ -79,12 +79,14 @@ process() {
 
   if [[ $file = training.*.gz ]]
   then
+    # check if the file already exists in the desired location
     if [ -f "$TESTDIR/$file" ] || [ -f "$TRAINDIR/$file" ]
     then
       echo -n "."
       return
     fi
 
+    # compute basic file integrity check
     size=$(zcat $dir/$file | wc -c)
     let rem="size % 8276"
     
@@ -94,6 +96,7 @@ process() {
       return
     fi
 
+    # new correct file, put in correct directory
     let "n++"
 
     id=$(echo $file | cut -d'.' -f 2)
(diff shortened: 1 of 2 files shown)


scripts/split.sh
```diff
@@ -46,19 +46,27 @@ do
...
+  echo "env var LC0LOCKFILE not set"
+  exit 1
+fi
 
 # clear test and train split dirs
-rm -rf $TESTDIR $TRAINDIR
-mkdir -vp $TESTDIR $TRAINDIR
+if [ ! -d "$TESTDIR" ] || [ ! -d "$TRAINDIR" ]
+then
+  rm -rf $TESTDIR $TRAINDIR
```
Tilps: Maybe quote around $TESTDIR and $TRAINDIR here - I'm always wary of doing an rm -rf without proper quoting.

scripts/split.sh
```diff
@@ -46,19 +46,27 @@ do
...
+  mkdir -vp $TESTDIR $TRAINDIR
+fi
 
-n=0
+let n="$(ls $TESTDIR | wc -l) + $(ls $TRAINDIR | wc -l)"
 let overhead="$WINSIZE / 10"
 let max="$WINSIZE + $overhead"
```
Tilps: I was thinking about this some more today - I think you probably should make this slightly larger than winsize + overhead - to have some slop in case the split isn't mathematically exact and the training script complains about not having enough chunks as a result.

Error323: Yeah I agree, it should never fall under the window size. So + some constant e.g. 5?

Tilps: I think 100 should be an obviously comfortable margin given the way the split is calculated - if we were using sha it would nee to be larger and probably would depend on the window size.

Error323: So winsize + overhead + 100?

Tilps: I think that should be good enough yes.

PR #18, opened 2018-06-10
Copy changed yaml file to github

borg323:
An untested suggestion to copy changed yaml file to a github branch every time training starts. If the yaml file is unchanged the git operation should not do anything.

### DIFF
--- a/tf/start.sh
+++ b/tf/start.sh
@@ -42,6 +42,10 @@ do
     # train all networks
     for netarch in ${NETARCHS[@]}
     do
+      echo "Saving .yaml changes:"
+      git add $CONFIGDIR/$netarch.yaml
+      git commit -m "Configuration change"
+      git push https://github.com/LeelaChessZero/lczero-training.git running_conf
       echo "Training $netarch:"
       train "$CONFIGDIR/$netarch.yaml" "${netarch}-$(date +"%Y_%m%d_%H%M_%S_%3N").txt"
     done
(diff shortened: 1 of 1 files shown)


tf/start.sh
```diff
@@ -42,6 +42,10 @@ do
     # train all networks
     for netarch in ${NETARCHS[@]}
     do
+      echo "Saving .yaml changes:"
+      git add $CONFIGDIR/$netarch.yaml
+      git commit -m "Configuration change"
+      git push https://github.com/LeelaChessZero/lczero-training.git running_conf
```
Error323: I think this is good indeed!
Probably the last command should be performed in parallel using `&`.

borg323: Done

tf/start.sh
```diff
@@ -42,6 +42,10 @@ do
     # train all networks
     for netarch in ${NETARCHS[@]}
     do
+      echo "Saving .yaml changes:"
+      git add $CONFIGDIR/$netarch.yaml
```
Tilps: This only works if configdir is under the git directory - which on test training is currently not true.

borg323: This is even better. Then we can do:
```
echo "Saving .yaml changes:"
pushd $CONFIGDIR
git add $netarch.yaml
git commit -m "Configuration change"
git push https://github.com/LeelaChessZero/lczero-training-conf.git &
popd
```
and have all changes in a new "lczero-training-conf" git repository

Error323: Sounds good to me!

PR #19, opened 2018-06-10
Optimize

Error323:
significantly speedup splitting functionality when the train/test split exists already.

### DIFF
--- a/scripts/diff.py
+++ b/scripts/diff.py
@@ -1,0 +1,36 @@
+#!/usr/bin/env python
+
+import glob
+import os
+import argparse
+
+
+def get_latest_chunk_ids(dirs):
+    chunks = []
+    for d in dirs:
+        for f in glob.glob(d + "/training.*.gz"):
+            chunks.append(int(os.path.basename(f).split('.')[-2]))
+    chunks.sort(reverse=True)
+    return chunks
+
+
+def main(argv):
+    a = get_latest_chunk_ids([argv.input])
+    n = min(argv.winsize, len(a))
+    b = get_latest_chunk_ids(argv.dirs)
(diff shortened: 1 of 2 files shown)


scripts/split.sh
```diff
@@ -79,30 +81,23 @@ process() {
 
   if [[ $file = training.*.gz ]]
   then
-    # check if the file already exists in the desired location
```
Tilps: Could leave this in place, just in case.

scripts/diff.py
```diff
@@ -0,0 +1,36 @@
...
+import os
+import argparse
+
+
+def get_latest_chunk_ids(dirs):
+    chunks = []
+    for d in dirs:
+        for f in glob.glob(os.path.join(d, "training.*.gz")):
+            chunks.append(int(os.path.basename(f).split('.')[-2]))
+    chunks.sort(reverse=True)
```
Tilps: Ahh, sorting by id rather than mtime - sounds good.

scripts/diff.py
```diff
@@ -0,0 +1,36 @@
...
+    return chunks
+
+
+def main(argv):
+    a = get_latest_chunk_ids([argv.input])
+    n = min(argv.wsize, len(a))
+    b = get_latest_chunk_ids(argv.dirs)
+    diff = set(a[:n]) - set(b)
+    for i in diff:
```
Tilps: I'm not sure about python set - does it iterate in sorted order from smallest to largest?

dubslow: Set iteration order is undefined, and in practice on CPython is hash dependent (read: reproducible within a given process, but ~random between processes).

PR #20, opened 2018-06-14
Save TensorBoard histogram summary of weights

Cyanogenoid:
The TensorBoard histogram feature allows visualisation of how the distribution of weights in the network changes over the training process ([Guide](https://www.tensorflow.org/programmers_guide/tensorboard_histograms)). This gives us more information for network debugging, similar and hopefully better than skewness measures that people have used.

This PR saves the histogram summary of the network weights necessary for this. TensorBoard will then pick them up automatically and add a Histogram tab where the weight histograms can be seen. This PR also names the weights and biases so that they are grouped correctly in TensorBoard ([Example](https://i.stack.imgur.com/IttNH.jpg) of what it may look like.)
This is currently untested, but scs-ben has agreed to check if it works soon.

### DIFF
--- a/tf/tfprocess.py
+++ b/tf/tfprocess.py
@@ -25,20 +25,20 @@ import bisect
 
 VERSION = 2
 
-def weight_variable(shape):
+def weight_variable(shape, name=None):
     """Xavier initialization"""
     stddev = np.sqrt(2.0 / (sum(shape)))
     initial = tf.truncated_normal(shape, stddev=stddev)
-    weights = tf.Variable(initial)
+    weights = tf.Variable(initial, name=name)
     tf.add_to_collection(tf.GraphKeys.REGULARIZATION_LOSSES, weights)
     return weights
 
 # Bias weights for layers not followed by BatchNorm
 # We do not regularlize biases, so they are not
 # added to the regularlizer collection
-def bias_variable(shape):
+def bias_variable(shape, name=None):
     initial = tf.constant(0.0, shape=shape)
-    return tf.Variable(initial)
+    return tf.Variable(initial, name=name)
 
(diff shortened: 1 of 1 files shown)


Cyanogenoid:
It seems to work fine on my machine and looks like this with a new net:
image
You can see the conv_weights being normally distributed as they should be and the batch norm params using the default initialisation. I haven't tested whether it keeps working when doing actual training, but I don't see why it wouldn't work then.

Note that because the variables have been renamed, old checkpoints can no longer be loaded easily. Without renaming the variables, the conv weights show up as Variable:0, Variable:1, etc. and not be correctly grouped into the bn0, bn1, etc. groups. Any merging of this PR should thus occur when starting a new net.

Error323:
Excellent! I'll merge it when we start a new test. Thanks!

Cyanogenoid:
Histograms are broken, it's not grouping them correctly together (note that there are beta_0, beta_0_1, beta_0_2, etc when there should be a single beta within bn0). I don't have computer access for a while, so the following changes need to be made to fix:
- Create the merged histogram nodes after init_net once instead of in every calculate_test_summary and save the merged summary node into self. Should just be moving the two lines that do the histogram stuff here-ish https://github.com/LeelaChessZero/lczero-training/blob/master/tf/tfprocess.py#L137 and changing the second line to only merge the histograms, without the .eval at the end.
- Put a tf.summary.merge back in calculate_test_summary to merge the scalars with the self.histograms and .eval it

This uses the same histogram node, so hopefully TB knows to keep those together. If this still doesn't work, tf.summary.scalar have to be used correctly instead of manually creating the test.Summary objects and filling them with values in order to avoid a tf.summary.merge every test step.

PR #22, opened 2018-06-22
track the LR

scs-ben:
This extensive PR forces the current value for Learning Rate to be added to the TensorBoard logs.
This change can be added to an existing network with no additional work having to be done.


Error323:
Excellent, I was about todo this! Thanks!

PR #24, opened 2018-06-26
Fix duplicate histogram summaries for same variable

Cyanogenoid:
Currently, there are 6 similar copies of each variable in the histogram tab: e.g. `bn0/batch_normalization/beta_0`, `.../beta_0_1`, ..., `.../beta_0_5`. I believe that this is due to new histogram nodes being created every time calculate_test_summaries is called, which causes TensorBoard to try to rename them.
This PR changes it so that histogram nodes are only created once, thus TB should be able to reuse the summary nodes to make the histograms easier to view.

Again, I don't have things set-up properly to be able to test whether this actually works (but I think it should). Merging this PR shouldn't need a newly generated checkpoint unlike last time https://github.com/LeelaChessZero/lczero-training/pull/20.


Cyanogenoid:
scs-ben confirmed that this fixes the histograms.

PR #25, opened 2018-07-03
grammatical fix

ganeshkrishnan1:
writted => written


dubslow:
probably "wrote" is even better

PR #29, opened 2018-07-06
Use ghost batch norm to match A0 paper

### LINKED ISSUE #28: Discrepancy of batch norm to A0 paper
In [A0](https://arxiv.org/abs/1712.01815), 64 TPU workers are used to train the network. That means that with their batch size of 4096, each worker independently processes a subset of every minibatch containing 64 samples. Thus, the batch statistics of the batch norms that each worker computes is actually over these 64 samples, not over the whole minibatch. Our current training code computes batch norm statistics over the whole minibatch.

In [this paper](https://arxiv.org/abs/1705.08741), "ghost batch normalisation" is discussed, which is a scheme equivalent to splitting the batch to multiple workers as was done in A0. They find that this improves generalisation performance (see Table 1 "+GBN" column), which should translate to better playing strength for Leela.

My suggestion is to change the training code to include ghost batch normalisation in order to match what is described in the A0 paper. Fortunately, this is simple to implement because [`tf.layers.batch_normalization`](https://www.tensorflow.org/api_docs/python/tf/layers/batch_normalization) supports it through the `virtual_batch_size` argument. In the three places in tfprocess.py where it is called, it simply needs another argument `virtual_batch_size=64` passed to it. No other changes should be needed.

Cyanogenoid:
Closes #28, see the discussion there for rationale. I don't think a consensus on whether this can be merged into the current test run or whether a new one should be started has been reached yet.

### DIFF
--- a/tf/tfprocess.py
+++ b/tf/tfprocess.py
@@ -377,6 +377,7 @@ class TFProcess:
                     conv2d(inputs, W_conv),
                     epsilon=1e-5, axis=1, fused=True,
                     center=True, scale=False,
+                    virtual_batch_size=64,
                     training=self.training)
         h_conv = tf.nn.relu(h_bn)
 
@@ -413,6 +414,7 @@ class TFProcess:
                     conv2d(inputs, W_conv_1),
                     epsilon=1e-5, axis=1, fused=True,
                     center=True, scale=False,
+                    virtual_batch_size=64,
                     training=self.training)
         h_out_1 = tf.nn.relu(h_bn1)
         with tf.variable_scope(weight_key_2):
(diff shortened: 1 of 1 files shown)


Cyanogenoid:
scs-ben is using this PR in his current no-history test and hasn't complained yet, so that may count as a test. Assuming that the TF implementation of ghost BN isn't bugged (I haven't seen any issue suggesting that it is), the BN weights won't be confused and the inference side with lc0 doesn't need to know about it either.

Tilps:
Maybe add a validation that the batch size is a multiple of the ghost batch size?  That is unless the error message that tensorflow is going to give in that scenario is trivial to map back to the cause, in which case I guess don't bother.

PR #30, opened 2018-07-06
Implement gradient accumulation

Cyanogenoid:
This PR allows training with larger batches than GPU memory can store at once. This is done by splitting a batch into many smaller batches, of which the gradients are accumulated to perform one gradient update. The implementation is adapted from [this StackOverflow answer](https://stackoverflow.com/questions/46772685/how-to-accumulate-gradients-in-tensorflow).

By default, no splitting of batches is performed. If `num_batch_splits` is added to the config yml under the training key, the batch is split into that many sub-batches which are then processed independently, accumulating gradients. This is very similar to training with the higher batch size in the first place, and it is *fully equivalent* if https://github.com/LeelaChessZero/lczero-training/pull/29 is merged.

As a result, this allows us to exactly match A0's training scheme which used a batch size of 4096 and LR of 0.2. The config to do this should contain
```yaml
training:
    batch_size: 4096
    num_batch_splits: 8
```
in order to have a total batch size of 4096, but perform this using 8 splits of this batch of size 512 each.

As a side-effect, the number of evaluation batches increases by a factor of `num_batch_splits` over actually using a batch size of 4096 because it is using 512 size batches for those. This is fine, since evaluation results do not depend on batch size. The averaging of evaluation results already takes care of this correctly.

### DIFF
--- a/tf/tfprocess.py
+++ b/tf/tfprocess.py
@@ -233,8 +233,10 @@ class TFProcess:
             self.avg_policy_loss.append(policy_loss)
             self.avg_mse_loss.append(mse_loss)
             self.avg_reg_term.append(reg_term)
+        # Gradients of batch splits are summed, not averaged like usual, so need to scale lr accordingly to correct for this.
+        corrected_lr = self.lr / batch_splits
         self.session.run(self.train_op,
-            feed_dict={self.learning_rate: self.lr / batch_splits, self.training: True, self.handle: self.train_handle})
+            feed_dict={self.learning_rate: corrected_lr, self.training: True, self.handle: self.train_handle})
 
         # Update steps since training should have incremented it.
         steps = tf.train.global_step(self.session, self.global_step)
--- a/tf/tfprocess.py
+++ b/tf/tfprocess.py
@@ -112,11 +112,16 @@ class TFProcess:
         opt_op = tf.train.MomentumOptimizer(
             learning_rate=self.learning_rate, momentum=0.9, use_nesterov=True)
 
+        # Accumulate (possibly multiple) gradient updates to simulate larger batch sizes than can be held in GPU memory.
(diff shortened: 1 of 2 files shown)


ganeshkrishnan1:
You would have to bootstrap the network with this config as the checkpoint does not have this new variable and restore fails

tf/train.py
```diff
@@ -130,11 +136,12 @@ def main(cmd):
         tfprocess.restore(cp)
 
     # Sweeps through all test chunks statistically
-	# Assumes average of 10 samples per test game.
+    # Assumes average of 10 samples per test game.
+    # Evaluation result doesn't depend on batch size, so we can use split instead of total batch size.
```
Tilps: This comment seems potentially confusing.  Maybe focus on the fact that test uses the same split batch size as train for simplicity, then mention that that is okay because test eval result is just an average which works out the same regardless of batch size.

tf/tfprocess.py
```diff
@@ -215,10 +220,21 @@ def process(self, batch_size, test_batches):
...
+                feed_dict={self.training: True, self.handle: self.train_handle})
+            # Keep running averages
+            # Google's paper scales MSE by 1/4 to a [0, 1] range, so do the same to
+            # get comparable values.
+            mse_loss /= 4.0
+            self.avg_policy_loss.append(policy_loss)
+            self.avg_mse_loss.append(mse_loss)
+            self.avg_reg_term.append(reg_term)
+        self.session.run(self.train_op,
+            feed_dict={self.learning_rate: self.lr / batch_splits, self.training: True, self.handle: self.train_handle})
```
Tilps: This is possibly worth a comment that self.lr / batch_splits is used because the gradient has been accumulated, not averaged.

PR #31, opened 2018-07-12
Protobuf

Error323:
Use protobuf instead of ascii textfiles for storing and distributing weights.

### DIFF
--- a/tf/net.py
+++ b/tf/net.py
@@ -1,4 +1,4 @@
-#!/usr/bin/env python
+#!/usr/bin/env python3
 
 import argparse
 import gzip
@@ -64,9 +64,9 @@ class Net:
             filename += ".txt.gz"
 
         with gzip.open(filename, 'wb') as f:
-            f.write("{}\n".format(2))
+            f.write("{}\n".format(2).encode('utf-8'))
             for row in weights:
-                f.write(" ".join(map(str, row.tolist())) + "\n")
+                f.write((" ".join(map(str, row.tolist())) + "\n").encode('utf-8'))
 
         size = os.path.getsize(filename) / 1024**2
         print("saved as '{}' {}M".format(filename, round(size, 2)))
--- a/tf/net.py
+++ b/tf/net.py
@@ -6,9 +6,9 @@ import os
 import numpy as np
 import lc0net_pb2 as pb
 
-LC0_MAJOR = 1
-LC0_MINOR = 0
-LC0_PATCH = 0
+LC0_MAJOR = 0
+LC0_MINOR = 14
+LC0_PATCH = 1
 
 
 class Net:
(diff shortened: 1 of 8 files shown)


tf/net.py
```diff
@@ -0,0 +1,190 @@
...
+        self.pb.min_version.major = LC0_MAJOR
+        self.pb.min_version.minor = LC0_MINOR
+        self.pb.min_version.patch = LC0_PATCH
+        self.weights = []
+
+
+    def fill_layer(self, layer, weights):
+        """Normalize and populate 8bit layer in protobuf"""
+        params = np.array(weights.pop(), dtype=np.float32)
+        layer.min_val = 0 if len(params) == 1 else np.min(params)
```
Tilps: This seems to be protecting against a divide by 0, but if the single parameter is 0, then you still get a divide by 0.
Probably better to just do max and min and if max == min, max = min + 1 or something?

tf/net.py
```diff
@@ -0,0 +1,190 @@
...
+
+    def denorm_layer(self, layer, weights):
+        """Denormalize a layer from protobuf"""
+        params = np.frombuffer(layer.params, np.uint16).astype(np.float32)
+        params /= 2**16-1
+        weights.insert(0, params * (layer.max_val - layer.min_val) + layer.min_val)
+
+
+    def denorm_conv_block(self, convblock, weights):
+        """Denormalize a convblock from protobuf"""
```
Tilps: This should be reversed since denorm all runs backwards right?

tf/net.py
```diff
@@ -0,0 +1,190 @@
...
+        layer.max_val = np.max(params)
+        params = (params - layer.min_val) / (layer.max_val - layer.min_val)
+        params *= 2**16-1
+        params = np.round(params)
+        layer.params = params.astype(np.uint16).tobytes()
+
+
+    def fill_conv_block(self, convblock, weights):
+        """Normalize and populate 8bit convblock in protobuf"""
+        self.fill_layer(convblock.weights, weights)
```
Tilps: I think this should also be reversed, since your fill logic below runs things backwards.

tf/tfprocess.py
```diff
@@ -279,6 +280,9 @@ def process(self, batch_size, test_batches, batch_splits=1):
                 tf.Summary.Value(tag="Reg term", simple_value=avg_reg_term),
                 tf.Summary.Value(tag="LR", simple_value=self.lr),
                 tf.Summary.Value(tag="MSE Loss", simple_value=avg_mse_loss)])
+            self.net.pb.training_params.learning_rate = self.lr
```
Tilps: Wouldn't it make more sense to include the test values for loss rather than the training values?

tf/net.py
```diff
@@ -0,0 +1,191 @@
+#!/usr/bin/env python
+
+import argparse
+import gzip
+import os
+import numpy as np
+import lc0net_pb2 as pb
+
+LC0_MAJOR = 0
+LC0_MINOR = 14
```
Tilps: I would think for actual submission, it would probably make sense to bump this to 15 (and correspondingly bump version in lc0).

tf/net.py
```diff
@@ -0,0 +1,191 @@
...
+            self.fill_conv_block(res.conv2, weights)
+            self.fill_conv_block(res.conv1, weights)
+
+        self.fill_conv_block(self.pb.weights.input, weights)
+
+
+def main(argv):
+    net = Net()
+
+    if argv.input.endswith(".txt"):
```
Tilps: Any thoughts on using auto detect here rather than file name extension?

tf/start.sh
```diff
@@ -88,7 +88,7 @@ do
     for netarch in ${NETARCHS[@]}
     do
       echo "Training $netarch:"
-      train "$CONFIGDIR/$netarch.yaml" "${netarch}-$(date +"%Y_%m%d_%H%M_%S_%3N").txt"
+      train "$CONFIGDIR/$netarch.yaml" "${netarch}-$(date +"%Y_%m%d_%H%M_%S_%3N")"
```
Tilps: train() needs updating since train.py now outputs .pb.gz for weights - but its going to try and gzip a non existant file and then try to move that result to net dir.

tf/net.py
Error323: we could, but this is solely for testing purposes so I don't think it's required. lc0 will support both types anyway.

tf/net.py
```diff
@@ -7,8 +7,8 @@
 import proto.net_pb2 as pb
 
 LC0_MAJOR = 0
-LC0_MINOR = 15
-LC0_PATCH = 0
+LC0_MINOR = 14
```
Tilps: I don't think we should expect LC0 version numbers to go backwards - if they have gone backwards, we should bump it ASAP.

PR #32, opened 2018-07-15
Add ability to use a different global step than 0 for net_to_model.py

Tilps:


Tilps:
For some reason my latest changes are not showing up in the diff yet - but I added command line parameter support rather than just being a hard coded constant.

Tilps:
Something has gone wrong with github, my branch has 2 commits but this PR refuses to pick it up.  Will close and recreate.

PR #33, opened 2018-07-15
Add ability to use a different global step than 0 for net_to_model.py

Tilps:


tf/net_to_model.py
```diff
@@ -74,5 +82,7 @@
 tfp.init_net(x)
 tfp.replace_weights(weights)
 path = os.path.join(os.getcwd(), cfg['name'])
-save_path = tfp.saver.save(tfp.session, path, global_step=0)
+update_global_step = tfp.global_step.assign(START_FROM)
+tfp.session.run(update_global_step)
+save_path = tfp.saver.save(tfp.session, path, global_step=START_FROM)
```
scs-ben: I don't think will set the step count to "START_FROM" in the actual files. The names will be right but TensorBoard will still start at 0. The line that sets the step count inside the files is in tfprocess.py.

Tilps: The session_run update global step is what makes START FROM actually be used.  I used this part of the script successfully on test training server to do the ghost norm transition.

PR #34, opened 2018-07-19
Switch to standard TF checkpoint file parsing rather than hand rolled.

Tilps:
untested, but it might make it work for windows.


Tilps:
ASilver tested and said it fixed his issues on windows training.

PR #37, opened 2018-07-26
Perform independent pruning of test and train to allow for rebalancing.

Tilps:
Untested.

### DIFF
--- a/scripts/split.sh
+++ b/scripts/split.sh
@@ -61,17 +61,21 @@ then
   mkdir -vp "$TESTDIR" "$TRAINDIR"
 fi
 
-let n="$(ls $TESTDIR | wc -l) + $(ls $TRAINDIR | wc -l)"
+let n_test="$(ls $TESTDIR | wc -l)"
+let n_train="$(ls $TRAINDIR | wc -l)"
+let n="$n_test + $n_train"
 let overhead="$WINSIZE / 10"
-let max="$WINSIZE + $overhead + 100"
+let max="$WINSIZE + $overhead + 200"
+max_train=$(echo "scale=1;($TRAINPCT / 100) * $max" | bc | cut -d'.' -f1)
+max_test=$(echo "scale=1;(1 - $TRAINPCT / 100) * $max" | bc | cut -d'.' -f1)
 overhead_train=$(echo "scale=1;($TRAINPCT / 100) * $overhead" | bc | cut -d'.' -f1)
 overhead_test=$(echo "scale=1;(1 - $TRAINPCT / 100) * $overhead" | bc | cut -d'.' -f1)
 
 echo ""
-echo "start splitter, found $n games"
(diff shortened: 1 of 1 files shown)


scripts/split.sh
```diff
@@ -61,17 +61,21 @@ then
...
 
-let n="$(ls $TESTDIR | wc -l) + $(ls $TRAINDIR | wc -l)"
+let n_test="$(ls $TESTDIR | wc -l)"
+let n_train="$(ls $TRAINDIR | wc -l)"
+let n="$n_test + $n_train"
 let overhead="$WINSIZE / 10"
-let max="$WINSIZE + $overhead + 100"
+let max="$WINSIZE + $overhead + 200"
+max_train=$(echo "scale=1;($TRAINPCT / 100) * $max" | bc | cut -d'.' -f1)
+max_test=$(echo "scale=1;(1 - $TRAINPCT / 100) * $max" | bc | cut -d'.' -f1)
```
dubslow: Couldn't this just be `let max_test="1 - $max_train"` ?

scripts/split.sh
```diff
@@ -61,17 +61,21 @@ then
...
 overhead_train=$(echo "scale=1;($TRAINPCT / 100) * $overhead" | bc | cut -d'.' -f1)
 overhead_test=$(echo "scale=1;(1 - $TRAINPCT / 100) * $overhead" | bc | cut -d'.' -f1)
```
dubslow: ditto here I guess

scripts/split.sh
```diff
@@ -99,26 +103,38 @@ process() {
...
+        ls -rt $TESTDIR | head -n $overhead_test | xargs -I{} rm -f $TESTDIR/{}
+        let "n -= $overhead_test"
+        let "n_test -= $overhead_test"
+        echo -n "-"
+      fi
+      if [ $n_train -gt $max_train ]
+      then
+        ls -rt $TRAINDIR | head -n $overhead_train | xargs -I{} rm -f $TRAINDIR/{}
+        let "n -= $overhead_train"
+        let "n_train -= $overhead_train"
```
dubslow: This style of `let` is inconsistent with the ones at the top, is that intended (target variable inside/outside the quotes)? Do they operate differently?

scripts/split.sh
Tilps: it would be max - max_train - but I'll leave that for a potential follow up.

scripts/split.sh
Tilps: Acknowledged.

scripts/split.sh
Tilps: Its because its a mutating assignment operator rather than just a straight assignment.

Tilps: So this code ended up being wrong after all - it was inside a subshell so the variable gets updated correctly, but only in the subshell.  The test machine is running a fixed version.

PR #38, opened 2018-08-01
multigpu

Error323:
This is quite an extensive rewrite of tfprocess.py. I ported the latest code from [leela-zero](https://github.com/gcp/leela-zero) to enable multigpu. We also get stochastic weight averaging for free as they implemented that also (very nice work).

For testing I trained both with the old code and new code. The new multigpu code seems to run 2x faster for 256x20 on my 2x 1080Ti. But going to 128x10 shows regression compared to the single gpu. Presumably there's too much cpu overhead.

### DIFF
--- a/tf/configs/example.yaml
+++ b/tf/configs/example.yaml
@@ -1,36 +1,57 @@
 %YAML 1.2
 ---
-name: 'kb1-64x6'                       # ideally no spaces
-gpu: 0                                 # gpu id to process on
+# name of the training project
+name: 'example'
 
 dataset: 
-  num_chunks: 100000                   # newest nof chunks to parse
-  train_ratio: 0.90                    # trainingset ratio
-  # For separated test and train data.
-  input_train: '/path/to/chunks/*/draw/' # supports glob
-  input_test: '/path/to/chunks/*/draw/'  # supports glob
-  # For a one-shot run with all data in one directory.
-  # input: '/path/to/chunks/*/draw/'
+    # newest nof chunks to parse
+    num_chunks: 120000
+    # trainingset ratio
+    train_ratio: 0.90
+    # For separated test and train data.
+    input_train: '/path/to/chunks/*/train/' # supports glob
+    input_test: '/path/to/chunks/*/test/' # supports glob
+    # For a one-shot run with all data in one directory.
+    # input: '/path/to/chunks/*/draw/' # supports glob
 
 training:
-    batch_size: 2048                   # training batch
-    test_steps: 2000                   # eval test set values after this many steps
-    train_avg_report_steps: 200        # training reports its average values after this many steps.
-    total_steps: 140000                # terminate after these steps
-    # checkpoint_steps: 10000          # optional frequency for checkpointing before finish
-    shuffle_size: 524288               # size of the shuffle buffer
-    lr_values:                         # list of learning rates
+    # gpu ids to train on
+    gpus:                              
+        - 0
+        - 1
+    # enable/disable stochastic weight averaging
+    swa: true
+    # training batch size
+    batch_size: 4096
+    # batchsize that fits in gpu memory, must be a multiple of batch_size
+    vram_batch_size: 512               
+    # eval test set values after this many steps
+    test_steps: 2000                   
+    # training reports its average values after this many steps.
+    info_steps: 200
+    # terminate after these steps
+    total_steps: 10000
+    # optional frequency for checkpointing before finish
+    checkpoint_steps: 2500
+    # size of the shuffle buffer
+    shuffle_size: 100000
+    # list of learning rates
+    lr_values:
+        - 0.2
         - 0.02
-        - 0.002
-        - 0.0005
-    lr_boundaries:                     # list of boundaries
-        - 100000
-        - 130000
-    policy_loss_weight: 1.0            # weight of policy loss
-    value_loss_weight: 1.0             # weight of value loss
-    path: '/path/to/store/networks'    # network storage dir
+    # list of boundaries
+    lr_boundaries:
+        - 5000
+    # weight of policy loss
+    policy_loss_weight: 1.0
+    # weight of value loss
+    value_loss_weight: 1.0
+    # network storage dir
+    path: '/path/to/store/networks'
 
 model:
-  filters: 64
-  residual_blocks: 6
+    # filters per layer
+    filters: 128
+    # residual blocks in resnet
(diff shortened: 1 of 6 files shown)


Cyanogenoid:
Do we want to port over Ghost BN (https://github.com/LeelaChessZero/lczero-training/pull/29) and Tensorboard histograms (https://github.com/LeelaChessZero/lczero-training/pull/20 https://github.com/LeelaChessZero/lczero-training/pull/24)?

I'm not sure if Ghost BN will still be useful since we are basically ensembling models with SWA and (at least from what I understand) Ghost BN can be considered a regularisation method, which tend to be less useful with ensembles.

Error323:
I'm not sure about ghost batching either. This implementation converges faster than what we had. I definitely would like the histograms back! They give nice insights! But to reduce complexity I preferred to have that in a separate pr.

tf/configs/example.yaml
```diff
@@ -1,36 +1,57 @@
...
-    lr_values:                         # list of learning rates
+    # gpu ids to train on
+    gpus:                              
+        - 0
+        - 1
+    # enable/disable stochastic weight averaging
+    swa: true
+    # training batch size
+    batch_size: 4096
+    # batchsize that fits in gpu memory, must be a multiple of batch_size
```
dubslow: er... shouldn't this read "must be a divisor of batch_size" :P

tf/tfprocess.py
```diff
@@ -16,34 +16,97 @@
...
 
 # Bias weights for layers not followed by BatchNorm
 # We do not regularlize biases, so they are not
 # added to the regularlizer collection
-def bias_variable(shape, name=None):
+def bias_variable(name, shape):
     initial = tf.constant(0.0, shape=shape)
-    return tf.Variable(initial, name=name)
+    bias = tf.get_variable(name, initializer=initial)
+    return bias
```
dubslow: Any reason these last two lines are split?

tf/tfprocess.py
```diff
@@ -16,34 +16,97 @@
...
+
+# Class holding statistics
+class Stats:
+    def __init__(self):
+        self.s = {}
+    def add(self, stat_dict):
+        for (k,v) in stat_dict.items():
+            if k not in self.s:
+                self.s[k] = []
+            self.s[k].append(v)
```
dubslow: Shouldn't this all just be 
```
def __init__(self):
    self.s = defaultdict(list)
```

and remove this `add` nonsense entirely?

tf/tfprocess.py
```diff
@@ -16,34 +16,97 @@
...
+        return np.mean(self.s[name] or [0])
+    def stddev_mean(self, name):
+        # standard deviation in the sample mean.
+        return math.sqrt(
+            np.var(self.s[name] or [0]) / max(0.0001, (len(self.s[name]) - 1)))
+    def str(self):
+        return ', '.join(
+            ["{}={:g}".format(k, np.mean(v or [0])) for k,v in self.s.items()])
+    def clear(self):
+        self.s = {}
```
dubslow: Duplicate code with `__init__`

tf/tfprocess.py
```diff
@@ -54,109 +117,244 @@ def __init__(self, cfg):
...
+
+        # Average gradients from different GPUs
+        self.loss = tf.reduce_mean(tower_loss)
+        self.policy_loss = tf.reduce_mean(tower_policy_loss)
+        self.mse_loss = tf.reduce_mean(tower_mse_loss)
+        self.reg_term = tf.reduce_mean(tower_reg_term)
+        self.y_conv = tf.concat(tower_y_conv, axis=0)
+        self.mean_grads = self.average_gradients(tower_grads)
+
+        # Do swa after we contruct the net
```
dubslow: typo

tf/tfprocess.py
```diff
@@ -54,109 +117,244 @@ def __init__(self, cfg):
...
+        if self.swa_enabled is True:
```
dubslow: `is True` in Python is essentially never the correct way to phrase it

ASilver:
A couple of minor issues to note:

- The logs are not being stored in the usual _tf/leelalogs_, they are being stored in _networks/networkname/logs_
- The step results are not displaying the LR values anymore. Ex: 
_Step 125, Policy=1.3567 mse=0.5789 etc_ 
instead of
_Step 125, LR=0.2 Policy=1.3567 etc_

ASilver:
I am testing the PR38 with SWA, and am getting very very strange values. I am testing 1 step per LR increment of 0.01, thus Step 11 is LR=0.1 and Step 12 is LR=0.11. Look at the crazy Policy values I am getting:

_step 12, policy=7.6417 mse=0.345703 reg=3.67696 total=12.7015 (1.02451 pos/s)
step 12, policy=1.23488e+06 training accuracy=1.10444%, mse=0.358436
step 13, policy=7.42069 mse=0.371094 reg=3.67696 total=12.582 (1.02643 pos/s)
step 13, policy=727089 training accuracy=1.09703%, mse=0.359836
step 14, policy=7.30655 mse=0.365723 reg=3.67696 total=12.4464 (1.03253 pos/s)
step 14, policy=446794 training accuracy=1.11984%, mse=0.359433
step 15, policy=7.57431 mse=0.349609 reg=3.67696 total=12.6497 (1.03233 pos/s)
step 15, policy=3.50153e+08 training accuracy=0.790655%, mse=0.35919_

tf/tfprocess.py
```diff
@@ -486,36 +607,130 @@ def construct_net(self, planes):
...
+        # save the current network.
+        self.snap_save()
+        # Copy the swa weights into the current network.
+        self.session.run(self.swa_load_op)
+        if self.swa_recalc_bn:
+            print("Refining SWA batch normalization")
+            for _ in range(200):
+                batch = next(data)
+                self.session.run(
+                    [self.loss, self.update_ops],
```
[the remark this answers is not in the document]
Cyanogenoid: Either way, as people have already mentioned in the Discord chat, I don't think this loop is necessary or even better than not adjusting BN params. It would be sensible if our architecture had Conv-ReLU-BN because the ReLU would affect the distribution of activations after SWA non-trivially, but our architecture with Conv-BN-ReLU means that BN params can be folded into the Conv layer easily, so I think not fine-tuning BN running mean and variances might actually the right thing to do.

Tilps:
We've got working multi-gpu support now, this PR is obsolete.

PR #40, opened 2018-08-12
Update TF requirement to 1.10

uyhcire:
Training crashes with 1.4 because it doesn't support `virtual_batch_size`


Error323:
It may not work for 1.4, but there are more versions between 1.4 and 1.10 which do work.

Tilps:
Obsolete since we've raised it to 1.12.2 per the automated security bot.

PR #43, opened 2018-08-30
Fix scale of weight initialization

Cyanogenoid:
The previous implementation of Xavier init is completely wrong for conv layers, initialising values too large. This is likely the cause for the long time that a new net needs to be trained before the mse loss drops, and as sergiovieri points out, during this period the value head output is always 1. Also, notice how the moving variances for the first BN in every residual block in the current Tensorboard are close to 1 at the start of the net as they should be, but gradually move to over 100 as you get deeper into the network. This is a sign that the scale of the activations in the identity path in the ResNet is growing significantly throughout residual blocks.

This PR changes the init to be the correct Xavier initialisation for a Gaussian (as opposed to the original uniform distribution in Xavier's paper, but this doesn't really matter). The implementation here matches the official TF implementation and the official PyTorch implementation:

https://github.com/tensorflow/tensorflow/blob/master/tensorflow/contrib/layers/python/layers/initializers.py#L62
https://github.com/pytorch/pytorch/blob/master/torch/nn/init.py#L225

For the conv layers, the stdev of the [3, 3, 256, 256] tensor is:

> sqrt(2 / (3 + 3 + 256 + 256)) = 0.062

when it should be

> sqrt(2 / (3 * 3 * 256 + 3 * 3 * 256)) = 0.021

in order to correctly maintain the scale of the activations and gradients throughout the network.

### DIFF
--- a/tf/tfprocess.py
+++ b/tf/tfprocess.py
@@ -34,7 +34,9 @@ def weight_variable(shape, name=None):
     else:
         fan_in = shape[0]
         fan_out = shape[1]
-    stddev = np.sqrt(2.0 / (fan_in + fan_out))
+    # truncated normal has lower stddev than a regular normal distribution, so need to correct for that 
+    trunc_correction = math.sqrt(1.3)
+    stddev = trunc_correction * np.sqrt(2.0 / (fan_in + fan_out))
     initial = tf.truncated_normal(shape, stddev=stddev)
     weights = tf.Variable(initial, name=name)
     tf.add_to_collection(tf.GraphKeys.REGULARIZATION_LOSSES, weights)
--- a/tf/tfprocess.py
+++ b/tf/tfprocess.py
@@ -27,7 +27,14 @@ from net import Net
 
 def weight_variable(shape, name=None):
     """Xavier initialization"""
-    stddev = np.sqrt(2.0 / (sum(shape)))
+    if len(shape) == 4:
+        receptive_field = shape[0] * shape[1]
+        fan_in = shape[2] * receptive_field
+        fan_out = shape[3] * receptive_field
+    else:
+        fan_in = shape[0]
+        fan_out = shape[1]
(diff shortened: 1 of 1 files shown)


Cyanogenoid:
After thinking about it, I'm not sure anymore if this actually fixes the aforementioned symptoms. The BNs should take care of most of the init issues already, so I'm starting to think that the main issue is the lack of gamma param in the BNs. Either way, this PR is a correctness fix, though it possibly won't have much visible effect.

sergiovieri:
Tested this,
```
step 1, policy=7.75473 training accuracy=0.0355114%, mse=0.240066
```
Initial mse looks really good and initial policy is not crazy 1e10

sergiovieri:
Tested this,
```
step 1, policy=7.75473 training accuracy=0.0355114%, mse=0.240066
step 50, lr=0.2 policy=6.67281 mse=0.477169 reg=0.534689 total=9.11617 (1056.34 pos/s)
step 100, lr=0.2 policy=5.99103 mse=0.479417 reg=0.548622 total=8.45731 (1938.11 pos/s)
step 100, policy=5.98204 training accuracy=4.79403%, mse=0.476829
```
policy does not start crazy 1e10

tf/tfprocess.py
```diff
@@ -27,7 +27,14 @@
...
     """Xavier initialization"""
-    stddev = np.sqrt(2.0 / (sum(shape)))
+    if len(shape) == 4:
+        receptive_field = shape[0] * shape[1]
+        fan_in = shape[2] * receptive_field
+        fan_out = shape[3] * receptive_field
+    else:
+        fan_in = shape[0]
+        fan_out = shape[1]
+    stddev = np.sqrt(2.0 / (fan_in + fan_out))
```
Tilps: Reading the tensorflow code you pointed at - it would appear that 2.6 instead of 2.0 here would be a closer match.

PR #45, opened 2018-09-05
Add support for warmup steps.

Tilps:
As a side effect of moving LR calculation before the training in each process call this also means that reported LR is actually the LR just used.  It also moves transitions between LR values one step earlier - but I think that is also fine.

I haven't tested this exact code, but I did test this strategy (with a value of approximately 130) and it, in combination with the recent changes to net initial weights, means we can successfully init a net without the MSE ever saturating.  So rather than up to 20k steps before we start reinforcement waiting for it to solve vanishing gradients, we are just teaching it the basics from the start pretty much.  I would suggest 1k steps would be plenty before releasing to reinforcement learning.

### DIFF
--- a/tf/tfprocess.py
+++ b/tf/tfprocess.py
@@ -115,6 +115,7 @@ class TFProcess:
 
         # Set adaptive learning rate during training
         self.cfg['training']['lr_boundaries'].sort()
+        self.warmup_steps = self.cfg['training'].get('warmup_steps', 0)
         self.lr = self.cfg['training']['lr_values'][0]
 
         # You need to change the learning rate here if you are training
@@ -236,6 +237,14 @@ class TFProcess:
             required_factor = 64 * self.cfg['training'].get('num_batch_splits', 1)
             raise ValueError('batch_size must be a multiple of {}'.format(required_factor))
 
+        # Determine learning rate
+        lr_values = self.cfg['training']['lr_values']
+        lr_boundaries = self.cfg['training']['lr_boundaries']
+        steps_total = steps % self.cfg['training']['total_steps']
+        self.lr = lr_values[bisect.bisect_right(lr_boundaries, steps_total)]
+        if self.warmup_steps > 0 and steps < self.warmup_steps
+             self.lr = self.lr * steps / self.warmup_steps
+
         # Run training for this batch
         self.session.run(self.zero_op)
         for _ in range(batch_splits):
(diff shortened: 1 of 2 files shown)


tf/tfprocess.py
```diff
@@ -258,12 +267,6 @@ def process(self, batch_size, test_batches, batch_splits=1):
         # Update steps since training should have incremented it.
         steps = tf.train.global_step(self.session, self.global_step)
 
-        # Determine learning rate
-        lr_values = self.cfg['training']['lr_values']
-        lr_boundaries = self.cfg['training']['lr_boundaries']
-        steps_total = (steps-1) % self.cfg['training']['total_steps']
```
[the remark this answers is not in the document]
Tilps: because this is after the step increment as training for the step has been run.  Thus global step counter may actually be a multiple of total_steps - and reporting that the LR just used is the first LR rather than the last LR would be pretty obviously wrong.

tf/tfprocess.py
```diff
@@ -120,6 +120,7 @@ def init_net(self, next_batch):
 
         # Set adaptive learning rate during training
         self.cfg['training']['lr_boundaries'].sort()
+        self.warmup_steps = self.cfg['training'].get('warmup_steps', 0)
         self.lr = self.cfg['training']['lr_values'][0]
```
Cyanogenoid: `self.lr = ...` can be removed here, since computing LR is done before the first training step now.

tf/tfprocess.py
```diff
@@ -268,6 +269,14 @@ def process(self, batch_size, test_batches, batch_splits=1):
             required_factor = 64 * self.cfg['training'].get('num_batch_splits', 1)
             raise ValueError('batch_size must be a multiple of {}'.format(required_factor))
 
+        # Determine learning rate
+        lr_values = self.cfg['training']['lr_values']
+        lr_boundaries = self.cfg['training']['lr_boundaries']
+        steps_total = steps % self.cfg['training']['total_steps']
+        self.lr = lr_values[bisect.bisect_right(lr_boundaries, steps_total)]
+        if self.warmup_steps > 0 and steps < self.warmup_steps:
+             self.lr = self.lr * steps / self.warmup_steps
```
Cyanogenoid: compute_update_ratio actually makes training crash because it doesn't like lr=0 at step=0 (all updates are 0 and are filtered away, then min of empty list is taken which crashes). `self.lr *= (steps + 1) / self.warmup_steps` should work.

Cyanogenoid:
Can confirm that this prevents the value head from saturating at the start of training on CCRL when setting value weight to 1 as well.

PR #46, opened 2018-10-01
WDL value head

Cyanogenoid:
This implements a Win/Draw/Loss value head with cross-entropy loss instead of the current scoring of wins as 1, draws as 0, and loss as -1 with an MSE loss. In early testing, it seems to be working.

MSE loss is no longer used in optimisation, but is still reported for comparison by averaging [1, 0, -1] weighted by the win, draw, and loss probabilities.

As this is a change in the architecture, merging this requires support in lc0.

### DIFF
--- a/tf/tfprocess.py
+++ b/tf/tfprocess.py
@@ -147,7 +147,7 @@ class TFProcess:
         self.policy_accuracy = tf.reduce_mean(correct_policy_prediction)
         correct_value_prediction = \
             tf.equal(tf.argmax(self.z_conv, 1), tf.argmax(self.z_, 1))
-        correct_value_prediction = tf.cast(correct_policy_prediction, tf.float32)
+        correct_value_prediction = tf.cast(correct_value_prediction, tf.float32)
         self.value_accuracy = tf.reduce_mean(correct_value_prediction)
 
         self.avg_policy_loss = []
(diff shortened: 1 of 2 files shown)


Cyanogenoid:
Interestingly, on the CCRL dataset, the final MSE loss is exactly the same as the baseline model that minimises MSE rather than cross-entropy on the value head. Policy loss is the same too. The value loss steadily goes down as you would expect.

Cyanogenoid:
I've renamed Accuracy to Policy Accuracy in Tensorboard for consistency in order to add Value Accuracy. Old logs are still compatible, they just still show up on Accuracy and thus are not as easily comparable.

tf/tfprocess.py
```diff
@@ -521,10 +543,10 @@ def construct_net(self, planes):
         self.weights.append(W_fc2)
         self.weights.append(b_fc2)
         h_fc2 = tf.nn.relu(tf.add(tf.matmul(h_conv_val_flat, W_fc2), b_fc2))
```
Tilps: I am wondering if this change should be accompanied by an increase in fc2 outputs to 256.  256 seems a small improvement for the old value head, but I wouldn't be surprised if its a stronger improvement for the new WDL head.

Cyanogenoid:
Superseded by https://github.com/LeelaChessZero/lczero-training/pull/63.

Issue #42, opened 2018-08-26
Add ratio of weights:updates to TensorBoard


cn4750:
See [here](http://cs231n.github.io/neural-networks-3/#ratio) for a description.
It seems that Minigo has [already added it](https://github.com/tensorflow/minigo/pull/63/files) to their TensorBoard.
This would help us determine when to drop the learning rate greatly.

PR #47, opened 2018-10-01
Log update ratio of weights

Cyanogenoid:
See #42. This also puts the update ratios into a histogram/distribution so that it's easier to view them across the whole network. Running mean and running variance are excluded since they are independent of learning rate.


Cyanogenoid:
Some conv weight ratios for a baseline net on CCRL at the start of training:

conv weight ratios for a baseline net on CCRL at the start of training

Tilps:
Any before and after performance metrics?

Cyanogenoid:
It only runs for steps where training stats are reported, so it shouldn't affect non-reporting steps at all. I made it run 500 reporting steps in a row: with this PR, it does 3.4 steps per second, without this PR it does 4.4 steps per second. test30 seems to be doing 1 reporting step per 125 steps, so this difference shrinks to next to nothing.

Tilps:
Yeap that sounds acceptable.

PR #50, opened 2018-10-17
SWA, attempt 2

Cyanogenoid:
This is a follow-up to https://github.com/LeelaChessZero/lczero-training/pull/38 with only the SWA parts pulled out and slightly refactored to be nicer. I took out the change to LR schedule for now as it isn't required for SWA (but the paper does say that it works well with cyclical learning rates).

A potential issue with Error's previous code was that swa_cycle (now renamed to swa_steps) -- how many steps between SWA updates there are -- was coupled with swa_max_n -- the maximum momentum term of the exponential moving average in the SWA. The SWA paper only updates the SWA weights at the end of each epoch (thousands or more steps) and has max_n at 40. Any suggestions for better names than swa_max_n, since it isn't a hard cut-off?

The original implementation was slow (less than half the positions/s than before on a 10b network) due to a `self.session.run(tf.assign_add(..., ...))` call having a massive overhead despite being a single operation. I avoided this by not keeping state of how many steps there are left until the next SWA update and instead just updating when `steps % <swa_steps> == 0`.

I removed the recalculating of the BN moving mean and variances for now, it increases time to create an SWA network a lot for probably not much benefit.  

The additional config needed is:
```yaml
     swa: True  # stochastic weight averaging
     swa_steps: 500  # frequency of updates to SWA network in iterations
     swa_max_n: 20  # limit momentum term to 1 - 1/swa_max_n after swa_max_n updates
```

### DIFF
--- a/tf/tfprocess.py
+++ b/tf/tfprocess.py
@@ -68,7 +68,7 @@ class TFProcess:
 
         self.swa_enabled = self.cfg['training']['swa']
 
-        # Limit momentum of SWA exponential average to 1 - 1/swa_max_n
+        # Limit momentum of SWA exponential average to 1 - 1/(swa_max_n + 1)
         self.swa_max_n = self.cfg['training']['swa_max_n']
 
         gpu_options = tf.GPUOptions(per_process_gpu_memory_fraction=0.90, allow_growth=True, visible_device_list="{}".format(self.cfg['gpu']))
--- a/tf/tfprocess.py
+++ b/tf/tfprocess.py
@@ -348,13 +348,15 @@ class TFProcess:
             swa_path = path + "-swa-" + str(steps)
             self.net.pb.training_params.training_steps = steps
             self.save_leelaz_weights(leela_path) 
-            self.save_swa_weights(swa_path)
             print("Weights saved in file: {}".format(leela_path))
-            print("SWA Weights saved in file: {}".format(swa_path))
+            if self.swa_enabled:
+                self.save_swa_weights(swa_path)
+                print("SWA Weights saved in file: {}".format(swa_path))
 
     def calculate_test_summaries(self, test_batches, steps):
-        self.snap_save()
-        self.session.run(self.swa_load_op)
+        if self.swa_enabled:
+            self.snap_save()
+            self.session.run(self.swa_load_op)
 
         sum_accuracy = 0
         sum_mse = 0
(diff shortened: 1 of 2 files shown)


tf/tfprocess.py
```diff
@@ -400,6 +441,38 @@ def log_histogram(self, tag, values, bins=1000):
...
+            save_ops = []
+            rest_ops = []
+            for var in self.weights:
+                if isinstance(var, str):
+                    var = tf.get_default_graph().get_tensor_by_name(var)
+                name = var.name.split(':')[0]
+                v = tf.Variable(var, name='save/'+name, trainable=False)
+                save_ops.append(tf.assign(v, var))
+                rest_ops.append(tf.assign(var, v))
+            self.save_op = tf.group(*save_ops)
```
Tilps: Maybe name these snap_save_op and snap_restore_op?

tf/tfprocess.py
```diff
@@ -66,6 +66,14 @@ def __init__(self, cfg):
         # For exporting
         self.weights = []
 
+        self.swa_enabled = self.cfg['training']['swa']
+
+        # Limit momentum of SWA exponential average to 1 - 1/swa_max_n
+        self.swa_max_n = self.cfg['training']['swa_max_n']
+
+        # Net sampling rate (e.g 2 == every 2nd network).
+        self.swa_c = self.cfg['training']['swa_steps']
```
Tilps: This seems unused.

Ttl: SWA skip has always been 1 in LZ as far as I know. I think it's pretty safe to remove this and assume that every test evaluation network is averaged.

Ttl:
The SWA paper does mention recalculating the BN moving averages. It was observed that SWA helped even without BN recalculation with LZ, but it would be nice to verify how big the difference is before removing it. LZ has a flag for enabling or disabling the BN recalculation, maybe it could be added to the PR?

Cyanogenoid:
I'll do the test with normal weights and SWA weights. I'll add a tf.summary.FileWriter for this.
I'll also check the influence of adapting the BN params. I doubt it has much of an effect on the nets that update SWA weights every step, but it might do something for the ones with less frequent updates.

Here are the results for steps=500, max_n=20 (swa) and steps=1, max_n=50 (swa_always):
2018-10-19-005908_809x548_scrot
2018-10-19-005927_831x551_scrot

PR #53, opened 2018-10-22
Add an upgrade script

Tilps:
This is a variant of update_steps that doesn't update steps, instead it uses code inspired by optimistic_restore from LZ training script.

Main differences from optimistic_restore is that this is very verbose about what data isn't being migrated between save files, so you might have some more confidence before training with the newly restore and saved checkpoint.

Note however that this script does not attempt to backup the checkpoint - it is assumed that if you screw up you've already made a backup, or you have a previous checkpoint that is close enough that you won't have lost too much training.

### DIFF
--- a/tf/upgrade.py
+++ b/tf/upgrade.py
@@ -1,0 +1,69 @@
+#!/usr/bin/env python3
+import argparse
+import os
+import yaml
+import sys
+import tensorflow as tf
+from tfprocess import TFProcess
+
+START_FROM = 0
+
+def main(cmd):
+    cfg = yaml.safe_load(cmd.cfg.read())
+    print(yaml.dump(cfg, default_flow_style=False))
+
+    root_dir = os.path.join(cfg['training']['path'], cfg['name'])
+    if not os.path.exists(root_dir):
+        os.makedirs(root_dir)
+
+    x = [
+        tf.placeholder(tf.float32, [None, 112, 8*8]),
+        tf.placeholder(tf.float32, [None, 1858]),
+        tf.placeholder(tf.float32, [None, 1])
+    ]
+
+    tfprocess = TFProcess(cfg)
+    tfprocess.init_net(x)
+
+    if os.path.exists(os.path.join(root_dir, 'checkpoint')):
+        cp = tf.train.latest_checkpoint(root_dir)
+        reader = tf.train.NewCheckpointReader(cp)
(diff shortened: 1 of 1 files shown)


tf/upgrade.py
```diff
@@ -0,0 +1,69 @@
...
+
+    tfprocess = TFProcess(cfg)
+    tfprocess.init_net(x)
+
+    if os.path.exists(os.path.join(root_dir, 'checkpoint')):
+        cp = tf.train.latest_checkpoint(root_dir)
+        reader = tf.train.NewCheckpointReader(cp)
+        saved_shapes = reader.get_variable_to_shape_map()
+        new_names = sorted(
+            [(var.name, var.name.split(':')[0]) for var in tf.global_variables()
```
Cyanogenoid: First element of this tuple doesn't seem to be used, other than being accessed but ignored in the loop below.

tf/upgrade.py
```diff
@@ -0,0 +1,69 @@
...
+             if var.name.split(':')[0] not in saved_shapes])
+        for var_name, saved_var_name in new_names:
+            print("New name {} will use default value ".format(saved_var_name))
```
Cyanogenoid: Extraneous space at the end.

tf/upgrade.py
```diff
@@ -0,0 +1,69 @@
...
+    x = [
+        tf.placeholder(tf.float32, [None, 112, 8*8]),
+        tf.placeholder(tf.float32, [None, 1858]),
+        tf.placeholder(tf.float32, [None, 1])
+    ]
+
+    tfprocess = TFProcess(cfg)
+    tfprocess.init_net(x)
+
+    if os.path.exists(os.path.join(root_dir, 'checkpoint')):
```
Cyanogenoid: Is there a use case for running this without starting from a checkpoint? Might be better to fail if no checkpoint found, otherwise it's not going to print anything about creating new names, implying that it upgraded the (non-existing) checkpoint correctly.

Tilps: good idea - done.

tf/upgrade.py
Tilps: yeap, I was just being lazy, cleaned it up.

tf/upgrade.py
Tilps: Fixed.

PR #55, opened 2018-10-23
Clip global norm of gradients

Cyanogenoid:
This should prevent the spikes we're experiencing in the RL pipeline if max_grad_norm is set appropriately. If it is not set, it simply logs the current global norm of the gradients into Tensorboard so that we can see what a reasonable value (slightly above typical values) for it can be. I checked that this prevents the value head from saturating on CCRL without the need for warmup.
This does not do anything if the measured norm is less than the specified norm, which is why defaulting this to an unreasonably large value makes it do nothing.


Cyanogenoid:
I confirmed that changing the configuration updates the value used for clipping after restoring from a checkpoint.

Tilps:
Yay, hopefully no more spikes.

PR #58, opened 2018-12-02
SE training code

Ttl:
Incompatible with the old network architecture. Shouldn't be too hard to add support for it, but I haven't bothered. Code for training the old networks is anyway available in the history if needed. `net.py` supports both network formats as it's needed for converting txt to protobuf.


Tilps:
This PR was missing an update to the libs/lczero-common submodule revision pin - will need one before the repo is usable 'as is' again.

PR #61, opened 2019-02-09
Support V4TrainingData

Cyanogenoid:
See https://github.com/LeelaChessZero/lc0/pull/722.
This is backwards compatible to loading V3TrainingData by setting root_q, best_q, root_d, root_q to 0.0 and not masking away illegal moves. Setting q_ratio in the interval [0, 1] (default 0) allows adjustment of how much q is mixed into z. In training, this PR changes nothing by default.

I started a training run on CCRL with v3 data and it hasn't broken at 1000 steps yet.

### DIFF
--- a/tf/chunkparser.py
+++ b/tf/chunkparser.py
@@ -26,8 +26,10 @@ import struct
 import tensorflow as tf
 import unittest
 
-VERSION = struct.pack('i', 4)
-STRUCT_STRING = '4s7432s832sBBBBBBBbff'
+V3_VERSION = struct.pack('i', 3)
+V4_VERSION = struct.pack('i', 4)
+V4_STRUCT_STRING = '4s7432s832sBBBBBBBbff'
+V3_STRUCT_STRING = '4s7432s832sBBBBBBBb'
 
 # Interface for a chunk data source.
 class ChunkDataSrc:
@@ -127,7 +129,8 @@ class ChunkParser:
             int8 result (1 byte)
             float32 q (4 bytes)
         """
-        self.v4_struct = struct.Struct(STRUCT_STRING)
+        self.v4_struct = struct.Struct(V4_STRUCT_STRING)
+        self.v3_struct = struct.Struct(V3_STRUCT_STRING)
 
 
     @staticmethod
@@ -164,6 +167,8 @@ class ChunkParser:
             uint8 rule50_count (1 byte)
             uint8 move_count (1 byte)
             int8 result (1 byte)
+            float32 root_q (4 bytes)
+            float32 best_q (4 bytes)
         """
         (ver, probs, planes, us_ooo, us_oo, them_ooo, them_oo, stm, rule50_count, move_count, winner, root_q, best_q) = self.v4_struct.unpack(content)
         # Enforce move_count to 0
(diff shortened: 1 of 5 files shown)


tf/chunkparser.py
```diff
@@ -113,7 +115,7 @@ def init_structs(self):
         struct.Struct doesn't pickle, so it needs to be separately
         constructed in workers.
 
-        V3 Format (8276 bytes total)
+        V3 Format (8292 bytes total)
```
Tilps: V4 format?

tf/decode_training.py
```diff
@@ -2095,11 +2103,18 @@ def main(args):
...
+                if version == VERSION3:
+                    record_size = V3_BYTES
+                else:
+                    record_size = V4_BYTES
+                for i in range(0, len(chunkdata), record_size):
+                    ts = TrainingStep(4 if version == VERSION4 else 3)
+                    record = chunkdata[i:i+record_size]
+                    if chunkdata[0:4] == VERSION3:
+                        record += 16 * b'\x00'
+                    ts.display_v2_or_v3(i//record_size, record)
```
Tilps: function name seems stale.

tf/tfprocess.py
```diff
@@ -95,18 +95,25 @@ def init_net(self, next_batch):
...
                                                     logits=self.y_conv)
         self.policy_loss = tf.reduce_mean(cross_entropy)
 
         # Loss on value head
+        q_ratio = self.cfg['training'].get('q_ratio', 0)
+        assert 0 <= q_ratio <= 1
+        target = self.q_ * q_ratio + self.z_ * (1 - q_ratio)
         self.mse_loss = \
-            tf.reduce_mean(tf.squared_difference(self.z_, self.z_conv))
+            tf.reduce_mean(tf.squared_difference(target, self.z_conv))
```
Tilps: I assume these changes don't cause any significant change in training speed. (I would hope not, but after SWA I like to ask...)

Cyanogenoid: It's just interpolating two tensors of size (batch_size, 1), should be completely negligible. Didn't seem slower when I ran this for 1k steps).

tf/decode_training.py
Cyanogenoid: v2 support doesn't seem to exist in chunkparser anymore, so removing that branch entirely.

tf/chunkparser.py
Cyanogenoid: Yep.

tf/chunkparser.py
```diff
@@ -310,20 +337,20 @@ def generate_fake_pos(self):
...
 
-    def v3_record(self, planes, i, probs, winner):
+    def v4_record(self, planes, i, probs, winner):
         pl = []
         for plane in planes:
             pl.append(np.packbits(plane))
         pl = np.array(pl).flatten().tobytes()
         pi = probs.tobytes()
-        return self.v3_struct.pack(VERSION, pi, pl, i[0], i[1], i[2], i[3], i[4], i[5], i[6], winner)
+        return self.v4_struct.pack(V4_VERSION, pi, pl, i[0], i[1], i[2], i[3], i[4], i[5], i[6], winner)
```
Tilps: Shouldn't this have some more fields added to its packing?

Cyanogenoid: Tests pass now.

tf/train.py
```diff
@@ -133,7 +133,7 @@ def main(cmd):
     # Assumes average of 10 samples per test game.
     # For simplicity, testing can use the split batch size instead of total batch size.
     # This does not affect results, because test results are simple averages that are independent of batch size.
-    num_evals = num_test*10 // ChunkParser.BATCH_SIZE
+    num_evals = num_test//10 // ChunkParser.BATCH_SIZE
```
Tilps: Please revert - then good to submit.

Tilps:
This has broken some of the supplementary scripts that create a dummy data entry in order to start init_net - they need to be updated for the additional passed parameter.

PR #63, opened 2019-02-10
Convolutional policy and WDL value head training code

Ttl:
WDL value head is mostly borrowed from earlier PR by @Cyanogenoid.

This also ended up including some other minor features. Number of test positions can be specified in yaml, some improvements to `net_to_model.py`. Restoring only restores matching variables allowing for example SWA to turned on and off during the training.

Conflicts with the latest training code changes for including Q. I don't have any chunks with Q in them, so it's hard to fix.

Due to protobuf changes generated nets needs minimum lc0 version of 0.21 even when not using any of the new features.

Tested by training net with convolutional policy and WDL head on CCRL dataset. It worked okay. Also trained some nets with old type heads for few steps, training looked okay and the nets worked in lc0.


Cyanogenoid:
This fails the tests in chunkparser, you might want to fix those.

I'd say that we can remove the `q_ratio` stuff in training for now, but keep `self.q_` available in `init_net` for a future PR that adds it back into the loss.

Cyanogenoid:
Note that current master supports loading both V3 and V4 training data, so existing data without Q can be loaded just fine, it's just always 0.0.

Ttl:
I made some changes to convert Q to WDL format for training. Untested since I don't have training data with Q, but the tests pass now and training without Q works.

tf/tfprocess.py
```diff
@@ -45,26 +48,81 @@ def weight_variable(shape, name=None):
...
 def bias_variable(shape, name=None):
     initial = tf.constant(0.0, shape=shape)
     return tf.Variable(initial, name=name)
 
+
 def conv2d(x, W):
     return tf.nn.conv2d(x, W, data_format='NCHW',
                         strides=[1, 1, 1, 1], padding='SAME')
 
+# Restore session from checkpoint. It silently ignore mis-matches
```
Tilps: Is there a reason to add optimistic_restore in this PR?  I prefer to use upgrade.py to avoid accidental upgrades in training.

tf/tfprocess.py
```diff
@@ -706,10 +847,11 @@ def construct_net(self, planes):
...
         self.weights.append(b_fc2)
         h_fc2 = tf.nn.relu(tf.add(tf.matmul(h_conv_val_flat, W_fc2), b_fc2))
-        W_fc3 = weight_variable([128, 1], name='fc3/weight')
-        b_fc3 = bias_variable([1], name='fc3/bias')
+        value_outputs = 3 if self.wdl else 1
+        W_fc3 = weight_variable([128, value_outputs], name='fc3/weight')
+        b_fc3 = bias_variable([value_outputs], name='fc3/bias')
         self.weights.append(W_fc3)
         self.weights.append(b_fc3)
-        h_fc3 = tf.nn.tanh(tf.add(tf.matmul(h_fc2, W_fc3), b_fc3), name='value_head')
```
Tilps: Shouldn't this be conditional on self.wdl - tanh if not wdl otherwise tanh?

Ttl: Yes, it should have tanh when not using WDL. Apparently it works close enough even without it and I didn't notice it in testing.

Tilps:
Please wait to submit until after I sort out why the current training code is broken :)

Ttl:
Do we want to merge this before 0.21 binary is released? Nets trained with this code don't work with older binaries due to protobuf changes.

Tilps:
Thats fine - I won't be upgrading the t40 training codebase once this is in.

PR #64, opened 2019-02-11
Train policy on legal moves only

Cyanogenoid:
This allows training to not compute a loss for illegal moves by setting the `mask_legal_moves` boolean option in the yaml training config. Instead of pushing the network to make illegal moves have 0 probability, it doesn't care about what probability it outputs for illegal moves. This is a sensible thing to do, since we don't care about what probability it puts on illegal moves in gameplay either. With the learned policy head, this gave about 30 Elo on CCRL. I haven't checked yet how much this gains with the A0 policy head, but I expect similar gains.


tf/tfprocess.py
```diff
@@ -135,10 +135,15 @@ def init_net(self, next_batch):
...
 
-        # y_ has -1 on illegal moves, flush them to 0 first
-        self.y_ = tf.nn.relu(self.y_)
-
         # Calculate loss on policy head
+        if self.cfg['training'].get('mask_legal_moves'):
+            # extract mask for legal moves from target policy
+            move_is_legal = tf.greater_equal(self.y_, 0)
+            # replace logits of illegal moves with large negative value (so that it doesn't affect policy of legal moves) without gradient
+            illegal_filler = tf.zeros_like(self.y_conv) - 1.0e-10
```
Ttl: 1e-10 is not very large value. I guess it should be 1e10?

Any possible numerical issues from using that large value? If Tensorflow calculates sum of policy, using large value for illegal moves can cause loss of precision.

Cyanogenoid: Whoops, indeed.

There shouldn't be any numerical issues. When the large negative value goes into the denominator of the softmax, it becomes virtually 0 because it appears as an exponent, so it doesn't affect the sum over exponentials in the denominator. When it appears in the nominator, it's multiplied by the 0 probability in the target vector when computing the loss, before the loss is reduced to a scalar, so again it doesn't affect anything.

tf/tfprocess.py
```diff
@@ -135,10 +135,15 @@ def init_net(self, next_batch):
         self.batch_norm_count = 0
         self.y_conv, self.z_conv = self.construct_net(self.x)
 
-        # y_ has -1 on illegal moves, flush them to 0 first
-        self.y_ = tf.nn.relu(self.y_)
-
         # Calculate loss on policy head
+        if self.cfg['training'].get('mask_legal_moves'):
+            # extract mask for legal moves from target policy
+            move_is_legal = tf.greater_equal(self.y_, 0)
```
Tilps: How confident are we that this works correctly with both y_ as negative and y_ as NaN?

Cyanogenoid: Quite confident: https://colab.research.google.com/drive/1jCLSgdkITC3rVN4qau6AQAZujAOvBsWo NaNs tend to compare false to everything, so this should be standard behaviour.
I don't know why our cross-entropy isn't NaN, it might be worth adding wrapping `self.y_` in a `tf.where(tf.is_nan(self.y_), tf.zeros_like(self.y_), self.y_)`

Tilps: I'm more confident that I'd notice if the cross entropy started calculating NaN in the presence of NaN training targets - so I think we can live with the current status there while we fix lc0 to not output NaN.

PR #67, opened 2019-03-07
Fix protobuf to txt conversion

Ttl:
Needed for parsing older networks without head format fields.

Also reduces the amount of decimals in txt to save some space.


Ttl:
Obsolete.

PR #68, opened 2019-03-12
net.py: Fix missing fields from network format

killerducky:
* Populate policyFormat and valueFormat fields in old protobufs
  without these fields.

* Allow saving .pb.gz back to another .pb.gz,
  useful if you add code to modify the pb weights.

* Print some more metadata about the protobuf file.

### DIFF
--- a/tf/net.py
+++ b/tf/net.py
@@ -221,9 +221,25 @@ class Net:
 
         return blocks // ws['residual']
 
+    def print_stats(self):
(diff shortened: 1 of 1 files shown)


killerducky:
Notes:
* I was accidentally pushing some branches to this repo instead of my own. I deleted them.
* Here is my messy branch that includes the gamma reduction code. See policy_gamma_correct(). https://github.com/killerducky/lczero-training/tree/nnpromohack

PR #70, opened 2019-03-16
Disable redundant gammas and add softmax feeding bias regularization.

Tilps:
I don't think these really need to be optional, so I didn't bother adding options.

### DIFF
--- a/tf/tfprocess.py
+++ b/tf/tfprocess.py
@@ -702,7 +702,7 @@ class TFProcess:
 
         gamma_key = weight_key + "/batch_normalization/gamma"
         if bn_scale:
-            gamma_key = gamm_key + ":0"
+            gamma_key = gamma_key + ":0"
         beta_key = weight_key + "/batch_normalization/beta:0"
         mean_key = weight_key + "/batch_normalization/moving_mean:0"
         var_key = weight_key + "/batch_normalization/moving_variance:0"
--- a/tf/tfprocess.py
+++ b/tf/tfprocess.py
@@ -687,7 +687,7 @@ class TFProcess:
                 virtual_batch_size=64,
                 training=self.training)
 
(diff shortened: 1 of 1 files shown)


Ttl:
I'm not really convinced that this is necessary and I don't think there is yet any convincing evidence either way for this.

The gammas after the first convolution in residual block and the ones in the heads are redundant and can be safely fixed to 1. They can merged through relu to the weights of the next layer. For that reason it feels weird to make them go to zero with regularization. If there is any possible issue with them I would rather fix them to 1 than apply regularization to them.

Tilps:
I also think fixing unneeded gammas would be better - I've not checked how easy that is to implement. (IE does the gamma weight go away entirely if I disable scale parameter.)

Tilps:
Disabling scale removes the variable - but I've kept it to keep the load/save code simple.  I also kept the name the same for consistency in the tensorboard.

Unfortunately net_to_model doesn't work for any old nets since they'll try and populate the unused gammas with non-unity values. Thus there is no way to migrate to this script from old versions.
I hope to write a gamma propagate function - but its far from trivial - so its not included here.

ASilver:
Does Leela Zero have this issue? I saw some ranking with a 256x20 SE net, but don't know if it was Leela, or another project.

Tilps:
Added gammas back to the initial convolution - since in the case input gammas would like to go to 0, the equivalent network with unit gammas has the gammas of residual block outputs going to infinity.  If there isn't a need for directly calculating 256 local values from the input they are forced to become a component of calculations that are not desired since actually getting the other gammas to go to infinity won't actually work.
This may imply that we should have some kind of regularization of the remaining gammas since they are still trivially infinite equal minimal loss solutions just by messing with the remaining gammas a little bit.

Tilps:
I'm merging this for T51.

PR #73, opened 2019-04-13
add "Total Train Loss" to tensorboard

aartdappel:
this was useful for me in a SL setting, to see if the net is optimizing anything to decide for a LR drop. I expect it to be extra useful in a RL setting where easier policy data is often harder for value and vice versa


Tilps:
I ended up adding this just recently, so this is now obsolete.

PR #76, opened 2019-05-12
Try and fix some memory leaks in saving.

Tilps:


tf/tfprocess.py
```diff
@@ -595,29 +595,31 @@ def save_swa_weights(self, filename):
...
-                #
-                work_weights = tf.transpose(weights, [1, 0])
-            else:
-                # Biases, batchnorm etc
-                work_weights = weights
-            all_evals.append(work_weights)
-        nparrays = self.session.run(all_evals)
+        if not hasattr(self, 'pb_save_op'):
+            all_evals = []
+            for e, weights in enumerate(self.weights):
```
Ttl: `e` is not used.

PR #78, opened 2019-06-15
Check best q and d values before potentially using them for training.

Tilps:
We've seen these values get corrupted, better to crash than use them incorrectly.

### DIFF
--- a/tf/chunkparser.py
+++ b/tf/chunkparser.py
@@ -202,7 +202,7 @@ class ChunkParser:
 
         best_q_w = 0.5 * (1.0 - best_d + best_q)
         best_q_l = 0.5 * (1.0 - best_d - best_q)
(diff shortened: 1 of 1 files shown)


tf/chunkparser.py
```diff
@@ -202,6 +202,7 @@ def convert_v4_to_tuple(self, content):
 
         best_q_w = 0.5 * (1.0 - best_d + best_q)
         best_q_l = 0.5 * (1.0 - best_d - best_q)
+        assert best_q >= -1.0 and best_d >= 0.0 and best_q <= 1.0 and best_d <= 1.0
```
mooskagh: In python it's possible to write `-1 <= best_q <= 1 and 0 <= best_d <= 1`.

Tilps: Done.

PR #81, opened 2019-07-18
ChunkParser is rewritten in order to support Python 3.7

Ghotrix:


Tilps:
A different approach ended up being taken, but the underlying problem is resolved. So this is obsolete.

PR #83, opened 2019-07-22
Some stuff from the training servers.

Tilps:


brianprichardson:
Perhaps it might be good to have a minimum number of chunks for the case when the path to the chunks is wrong and it does not find any, or starts training with far too few.

Tilps:
It might, but its value isn't high enough for me to go and add it right now.

PR #86, opened 2019-10-08
Update requirements.txt

aartdappel:


Tilps:
Closing in preference to dependabot - scripts are still actively used on 1.13, so raising it to 1.14 seemed less good than just raising it to the minimum needed to suppress the warning.

PR #88, opened 2019-11-13
updated yaml

aartdappel:


Ttl:
I think we should have explanations of all of the options in the example yaml file. Currently most of the options are not explained anywhere. Also there are some limitations on number of filters and se_ratios that are faster on cudnn-fp16 backend that could be mentioned here.

PR #91, opened 2019-12-07
Migrate to tf 1.14

Tilps:
This is intended as a stepping stone towards migrating to 2.0


Tilps:
Obsolete due to migration to tf2.

PR #92, opened 2019-12-07
Migrate to tf 2.0

Tilps:
This 'works' - but is extremely far from acceptable '2.0' code. I'm working on follow up PRs to make it more 'idiomatic' where possible...


Tilps:
or maybe it doesn't work ... reg term looks silly...

Tilps:
Obsolete by the real migrate to tf2.

PR #93, opened 2019-12-08
Actually migrate to tf 2.0

Tilps:
This is very much WIP, many many things unimplemented right now.


Tilps:
This is almost done now.  Classical policy head network save/load still pending, but otherwise it should be a feature match for before tf2 conversion.

tf/chunkparser.py
```diff
@@ -426,7 +426,7 @@ def test_tensorflow_parsing(self):
         best_q = best_q.reshape(batch_size, 3)
 
         # Pass it through tensorflow
-        with tf.Session() as sess:
+        with tf.compat.v1.Session() as sess:
```
Tilps: I didn't convert this test.

tf/tfprocess.py
```diff
@@ -128,191 +113,167 @@ def __init__(self, cfg):
...
+            self.swa_writer = tf.summary.create_file_writer(
                 os.path.join(os.getcwd(), "leelalogs/{}-swa-test".format(self.cfg['name'])))
-        self.histograms = [tf.summary.histogram(
-            weight.name, weight) for weight in self.weights]
-
-        self.init = tf.global_variables_initializer()
-        self.saver = tf.train.Saver()
-
-        self.session.run(self.init)
+        self.checkpoint = tf.train.Checkpoint(optimizer=self.orig_optimizer, model=self.model, global_step=self.global_step, swa_count=self.swa_count)
```
cn4750: This line fails when `swa: false` is set since it is set to None and isn't a trackable object.

tf/train.py
```diff
@@ -25,6 +25,7 @@
 import random
 import multiprocessing as mp
 import tensorflow as tf
+#tf.compat.v1.disable_v2_behavior()
```
Ttl: This can be removed

tf/tfprocess.py
```diff
@@ -523,134 +547,113 @@ def calculate_test_summaries(self, test_batches, steps):
...
+        permuted_tensors[-7] = all_tensors[-13]
+        permuted_tensors[-8] = all_tensors[-14]
+        permuted_tensors[-9] = all_tensors[-16]
+        permuted_tensors[-10] = all_tensors[-5]
+        permuted_tensors[-11] = all_tensors[-6]        
+        permuted_tensors[-12] = all_tensors[-7]        
+        permuted_tensors[-13] = all_tensors[-8]        
+        permuted_tensors[-14] = all_tensors[-9]        
+        permuted_tensors[-15] = all_tensors[-10]        
+        permuted_tensors[-16] = all_tensors[-15]
```
Ttl: There is lot of trailing whitespace here.

tf/tfprocess.py
```diff
@@ -523,134 +547,113 @@ def calculate_test_summaries(self, test_batches, steps):
...
+                # Biases, batchnorm etc
+                # pb expects every batch norm to have gammas, but not all of our
+                # batch norms have gammas, so manually add pretend gammas.
+                if 'beta:' in weights.name and not last_was_gamma:
+                    all_tensors.append(tf.ones_like(weights))
+                work_weights = weights.read_value()
+            all_tensors.append(work_weights)
+            last_was_gamma = 'gamma:' in weights.name
+
+        # HACK: model weights ordering is some kind of breadth first traversal,
```
Ttl: This looks too fragile. Should probably rework the `net.py` to assign the weights in proto based on the tensor names instead. I'll try to look into if that looks doable.

tf/tfprocess.py
Ttl: Works when `self.swa_count` definition above is taken out of the if clause.

tf/tfprocess.py
```diff
@@ -323,45 +284,90 @@ def replace_weights(self, new_weights):
...
+    @tf.function()
+    def read_weights(self):
+        return [w.read_value() for w in self.model.weights]
+
+    @tf.function()
+    def process_inner_loop(self):
+        print('tracing inner loop!')
+        x, y, z, q = next(self.train_iter)
+        with tf.GradientTape() as tape:
+            policy, value = self.model(x)
```
Ttl: `self.model` call should have `training=True` to set batch norm to training mode.

tf/tfprocess.py
```diff
@@ -439,76 +451,88 @@ def process(self, batch_size, test_batches, batch_splits=1):
...
         self.test_writer = true_test_writer
-        self.snap_restore()
+        for (old, w) in zip(backup, self.model.weights):
+            w.assign(old)
+
+    @tf.function()
+    def calculate_test_summaries_inner_loop(self):
+        print('tracing summaries inner loop!')
+        x, y, z, q = next(self.test_iter)
+        policy, value = self.model(x)
```
Ttl: `training=False` here.

Ttl:
With batch norm fixes I can restore old training using `net_to_model.py` with losses staying the same.

Tilps:
Thanks for the catch on the batch norm not being right - all the examples of custom training loops in the tf2 documentation I found omit setting the training flag, but I've reviewed the tf code for keras model test_on_batch/train_on_batch in combination with graph model (like the kind we are using) and it seems they always explicitly pass the training flag as you suggest. So that is fixed.

tf/tfprocess.py
Tilps: Done.

tf/tfprocess.py
```diff
@@ -323,45 +284,90 @@ def replace_weights(self, new_weights):
...
-            self.process(batch_size, test_batches, batch_splits=batch_splits)
+            self.process_v2(batch_size, test_batches, batch_splits=batch_splits)
+
+    @tf.function()
+    def read_weights(self):
+        return [w.read_value() for w in self.model.weights]
+
+    @tf.function()
+    def process_inner_loop(self):
+        print('tracing inner loop!')
```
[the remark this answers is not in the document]
Tilps: Done.

tf/train.py
Tilps: Done.

tf/tfprocess.py
```diff
@@ -439,76 +451,88 @@ def process(self, batch_size, test_batches, batch_splits=1):
...
-        self.calculate_test_summaries(test_batches, steps)
+        self.calculate_test_summaries_v2(test_batches, steps)
         self.test_writer = true_test_writer
-        self.snap_restore()
+        for (old, w) in zip(backup, self.model.weights):
+            w.assign(old)
+
+    @tf.function()
+    def calculate_test_summaries_inner_loop(self):
+        print('tracing summaries inner loop!')
```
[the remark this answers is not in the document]
Tilps: Done.

tf/tfprocess.py
Tilps: I agree, but it was a lot easier to do this way than to do properly, so I'm going to stick with it for right now. If you want to give doing it right a shot, I'm happy :)

tf/tfprocess.py
Tilps: I think I fixed it...

tf/tfprocess.py
Tilps: Indeed - fixed.

tf/tfprocess.py
Tilps: Yeap.

Tilps:
Going to submit with the experimental data loader unreviewed - its experimental.
Will follow up with more optimization and tweaks.

PR #94, opened 2019-12-13
Some micro-optimizations for tf2.

Tilps:
Some cleanup for swa related functions and use @tf.function on more stuff.


Tilps:
Accidentally reused same branch - the diff is small, ignore the commit history :P

Tilps:
I should probably work out if any of these are still useful...

PR #96, opened 2019-12-14
Save weights file with run name in the filename

Ttl:
TF 2.0 changes removed the run name from the saved weight file filename. This restores the old behaviour.

Also sneaked in a commit to make the number of workers configurable, with high core CPUs the default behaviour makes very many workers which uses a lot of memory.

### DIFF
--- a/tf/tfprocess.py
+++ b/tf/tfprocess.py
@@ -139,7 +139,7 @@ class TFProcess:
         if self.swa_enabled:
             # Count of networks accumulated into SWA
             self.swa_weights = [tf.Variable(w, trainable=False) for w in self.model.weights]
-        
+
         self.active_lr = 0.01
         self.optimizer = tf.keras.optimizers.SGD(learning_rate=lambda: self.active_lr, momentum=0.9, nesterov=True)
         self.orig_optimizer = self.optimizer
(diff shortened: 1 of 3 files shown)


tf/tfprocess.py
```diff
@@ -487,14 +487,13 @@ def process_v2(self, batch_size, test_batches, batch_splits=1):
             self.manager.save()
             print("Model saved in file: {}".format(self.manager.latest_checkpoint))
             evaled_steps = steps.numpy()
-            leela_path = self.manager.latest_checkpoint + "-" + str(evaled_steps)
-            swa_path = self.manager.latest_checkpoint + "-swa-" + str(evaled_steps)
+            path = os.path.join(self.root_dir, self.cfg['name'])
```
Tilps: I don't see how this does what is intended. It would appear to me that while the run name is in the filename, its no longer in the path as well, so these weights files will be saved up one directory from where they used to be.

Tilps: Nevermind, I was too sleepy to recall that root_dir already contains the name.

However since checkpoints don't contain step count in their filename anymore - I think this is still a bit awkward to map checkpoint to file.  I think I would rather perform some other mutation of latest_checkpoint to add the run as part of the name.

Tilps: Changing the checkpoint filename seems a reasonable alternative - so resolved!

Ttl:
The latest commit also adds the run name to checkpoint filename. Now both checkpoint and weight filenames should be the same as before tf 2.0 upgrade.

PR #97, opened 2019-12-14
Randomly flip training positions

Ttl:
When castling is not possible randomly flip the board including policy probabilities.

Currently this seems to result in a weaker net:
```
   # PLAYER              :  RATING  ERROR  POINTS  PLAYED   (%)  CFS(%)    W    D    L  D(%)
   1 lc0_no_flip_200k    :     0.0    7.9   530.5    1000    53     100  325  411  264    41
   2 lc0_flip_200k       :   -21.5    7.9   469.5    1000    47     ---  264  411  325    41
fixed 800 nodes.
```

Since castling is determined from the current position history may contain positions where king is on the wrong square. Some positions are not equally likely to be encountered in match play in a flipped state and it's probably not useful to train NN on those cases. This probably makes more sense when training with Fischer random training data. For ordinary chess more care should be used to determine if position can be flipped.

### DIFF
--- a/tf/chunkparser.py
+++ b/tf/chunkparser.py
@@ -25,6 +25,7 @@ import shufflebuffer as sb
 import struct
 import tensorflow as tf
 import unittest
+import lc0_az_policy_map
 
 V4_VERSION = struct.pack('i', 4)
 V3_VERSION = struct.pack('i', 3)
@@ -41,6 +42,11 @@ class ChunkDataSrc:
         return self.items.pop()
 
 
+def flip_vertex(v):
+    c = v % 8
+    r = v // 8
+    c = 7 - c
+    return 8 * r + c
 
 class ChunkParser:
     # static batch size
@@ -67,6 +73,15 @@ class ChunkParser:
         TensorFlow doesn't have a fast way to unpack bit vectors. 7950 bytes
         long.
         """
+
+        self.policy_flip_map = lc0_az_policy_map.make_map(kind='flip_permutation')
+
+        # Build full flip tables for flipping all input planes at once.
+        self.full_flip_map = np.array([flip_vertex(vertex) + p*8*8
+                for p in range(104) for vertex in range(8*8)], dtype=np.int32)
+
+        r = np.array(list(range(104*8*8)), dtype=np.int32)
+        assert np.array_equal((r[self.full_flip_map])[self.full_flip_map], r)
 
         # Build 2 flat float32 planes with values 0,1
         self.flat_planes = []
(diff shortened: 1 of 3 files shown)


Ttl:
I trained NN to predict if a chess position is flipped or not using the flipping logic in this PR. 41% of the positions were flipped and the network's success rate in predicting if the position is flipped was 89%. It looks like most of the flipped positions are clearly different enough from regular positions.

Ttl:
The latest commit flips position with 50% probability if there are less than 16 pieces and both kings are not in the starting or castling squares in the oldest move in the history. This gets rid of opposite castling moves and positions that can be very easily determined to be flipped. With this change about 40% of the positions can be flipped on T40 data.

There were some strange positions on T40 data with nans on Q and positions without kings, the latest commit also replaces assert with dropping the bad positions to avoid crashing on them.

Training NN to predict if the positions are flipped or not gets about the same 90% correct with 20% of the positions flipped, so it can still predict if the position is flipped with better than random chance. The net trained with flipping is little stronger on my tests, but it could be within the random variation:

```
800 nodes:

   # PLAYER         :  RATING  ERROR  POINTS  PLAYED   (%)  CFS(%)    W    D    L  D(%)
   1 lc0_flip       :     7.8    7.9   511.0    1000    51      83  301  420  279    42
   2 lc0_no_flip    :     0.0    7.9   489.0    1000    49     ---  279  420  301    42

White advantage = 40.15 +/- 8.04
Draw rate (equal opponents) = 42.50 % +/- 1.53

1600 nodes:

   # PLAYER         :  RATING  ERROR  POINTS  PLAYED   (%)  CFS(%)    W    D    L  D(%)
   1 lc0_flip       :    17.2    7.2   524.5    1000    52      99  272  505  223    50
   2 lc0_no_flip    :     0.0    7.2   475.5    1000    48     ---  223  505  272    50

White advantage = 5.27 +/- 7.62
Draw rate (equal opponents) = 50.63 % +/- 1.59

TC: 1+0.2. SF 1 thread. RTX 2080

   # PLAYER         :  RATING  ERROR  POINTS  PLAYED   (%)  CFS(%)    W    D    L  D(%)
   1 lc0_flip       :    37.4   19.4   221.0     400    55      70  144  154  102    38
   2 lc0_no_flip    :    27.6   19.5   215.5     400    54      98  129  173   98    43
   3 sf_dev         :     0.0   11.2   363.5     800    45     ---  200  327  273    41

White advantage = 39.51 +/- 9.13
Draw rate (equal opponents) = 41.62 % +/- 1.74
```

The test losses without flipping are similar on both nets trained with and without flipping. Before the test losses seemed to be worse with flipping.

This still defaults off. Add `flip_augmentation: True` under `dataset` in yaml to turn it on.

Tilps:
This should become obsolete if the migration to input format 4 is successful, but that is still pending.

PR #100, opened 2020-01-02
Fix net_to_model to save with checkpoint number

sergiovieri:


Tilps:
While you are here, you could make the same change in update_steps.py

PR #101, opened 2020-01-05
Add optional validation dataset

Tilps:
Every validation_steps or total_steps the entire validation dataset is enumerated.  No shuffling no sampling.
Not currently sorted, so if validation dataset is not an exact multiple of split batch size there could be some run to run variance.
Only supports v4 data.


Tilps:
Merging as trivial, will follow up on any feedback post submit.

PR #103, opened 2020-02-21
Direct to protobuf save and replace

Ttl:
Removes the step where weights are stored in list and instead writes them directly to protobuf based on Tensorflow variable names. Fixes classical policy head support.

Tested successfully shortly training from scratch and `net_to_model` from existing network with and without renorm.

Also some other formatting fixes. Removing backslashes fixes some graph optimizer error messages. There was a tensorflow bug report about it, but I can't find it right now.

### DIFF
--- a/tf/net.py
+++ b/tf/net.py
@@ -11,6 +11,11 @@ LC0_MINOR = 21
 LC0_PATCH = 0
 WEIGHTS_MAGIC = 0x1c0
 
+def nested_getattr(obj, attr):
+    attributes = attr.split(".")
+    for a in attributes:
+        obj = getattr(obj, a)
+    return obj
 
 class Net:
     def __init__(self,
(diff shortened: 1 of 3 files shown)


Tilps:
Thanks for this work.

I have concerns that the weight names will have an additional prefix under some circumstances - possibly mixed precision enabled, and in my multi-gpu branch if I ever manage to get it to actually be faster.  There are some exact match checks for the name of the first layer, and the name split is indexed by 0 in at least one case.  Could we either detect a prefix if it exists (using the first layer as clue) and strip it from all items before split or equality check, or make all names have the same level of / nesting so we can consistently reference using negative indexes for all cases of name split.  Neither option seems great, but the former is probably easiest I guess.

Ttl:
Half precision did add prefix in TF1.x, but names are identical in TF 2.0. I don't think the change would be difficult, but I would rather leave is as a TODO for when it's needed. It's easier to test when there actually are prefixes.

Tilps:
Sure, if you've tested half precision - lets go ahead with a TODO.

Ttl:
Yes, I tested half precision and it works.

tf/net.py
```diff
@@ -178,6 +201,113 @@ def save_proto(self, filename):
...
+
+        return (pb_name, block)
+
+    def get_weights_v2(self, names):
+        # `names` is a list of Tensorflow tensor names to get from the protobuf.
+        # Returns list of [Tensor name, Tensor weights].
+        tensors = {}
+
+        for tf_name in names:
+
```
Tilps: This newline here seems out of place.

PR #104, opened 2020-02-21
Less memory intensive input pipeline

Ttl:
The current input data pipeline stores a copy of list of chunks for every reader process. Since by default there is only one game per chunk, with millions of chunks the list of filenames can take hundreds of MB of memory. When that is multiplied by number of reader processes it can take many GB.

This PR modifies the pipeline so that the list of files is stored only once and changes the multiprocessing start method from spawn to platform default (fork on Linux, spawn on Windows). fork start_method can reuse shared read only memory for each process further decreasing the memory consumption. I left it commented out for testing. There are some things that work differently with spawn and fork methods.

Script is included for shuffling and merging training data on disk. Using much larger chunks makes it possible to train from hard disk.

Also slightly unrelated change of making `virtual_batch_size` configurable in yaml. In my tests `virtual_batch_size` slightly loses strength and is slower to train. It wasn't tested when it was merged originally.

Some tests:

```
128x10 net with 2048 batch size, 1 split, half precision. RTX2080

SKIP, pipeline, virtual_batch_size, pos/s, GPU usage, memory
32, old, 64, -, -, OOM (>16G)
32, new (spawn), 64, 5531, 90, 13.9G
32, new (fork), 64, 5537, 90, 11.0G
32, new (fork), None, 6272, 60, 11.0G
1, new (fork), None, 8089, 80, 11.0G
```

This isn't still fast enough to fully use GPU on my PC. I think it's the single threaded `parse` function that is limiting the performance. Combining the chunk reading part of this and parsing part of the `experimental_v4_only_dataset` should result in much faster input pipeline.

I tested this on Linux, but it still needs testing on Windows before I'm comfortable with merging. @jjoshua2 ?

### DIFF
--- a/scripts/shuffle.py
+++ b/scripts/shuffle.py
@@ -1,0 +1,64 @@
+#!/usr/bin/python3
+import gzip
+import sys
+import glob
+import os
+import random
+from multiprocessing import Pool
+import tqdm
+
+merge_files = 100
+processes = 8
+shuffle = True
+record_length = 8292
+
+def split(a, n):
+    k, m = divmod(len(a), n)
+    return [a[i * k + min(i, m):(i + 1) * k + min(i + 1, m)] for i in range(n)]
+
+def positions(chunk):
+    pos = []
+    i = 0
+    while record_length*i < len(chunk):
+        pos.append(chunk[record_length*i:record_length*(i+1)])
+        i += 1
+    return pos
+
+def shuffle(files):
+    data = []
+    for filename in files:
+        with gzip.open(filename, 'rb') as f:
+            data.extend(positions(f.read()))
+    if shuffle:
+        random.shuffle(data)
+    for d in data:
+        if d[0] != 0x04:
+            print(files)
+            raise ValueError('Wrong training data format, not V4')
+    new_file = list(os.path.splitext(files[0]))
+    new_file[0] += '_shuffled'
+    new_file = ''.join(new_file)
+    new_file_temp = new_file + '.temp'
+    with gzip.open(new_file_temp, 'wb', compresslevel=9) as f:
+        for d in data:
+            f.write(d)
+    # For interrupt safety, make sure not to write partial chunks.
+    os.rename(new_file_temp, new_file)
+    for filename in files:
+        os.remove(filename)
+
+if __name__ == "__main__":
+    if len(sys.argv) != 2:
+        print('Expected one argument, got {}'.format(len(sys.argv)-1))
+        exit(1)
+    s = sys.argv[1]
+    files = glob.glob(os.path.join(sys.argv[1], '*.gz'))
+    files = [f for f in files if '_shuffled' not in f]
+    print('Found {} files'.format(len(files)))
+    if len(files) == 0:
+        exit(1)
+    files = split(files, len(files)//merge_files)
+    pool = Pool(processes)
+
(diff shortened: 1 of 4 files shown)


cn4750:
This unfortunately doesn't seem to work on Windows with newer version of Python for the same reason the old training code did not run on Windows. Running Python 3.7 results in the following error message:
```Traceback (most recent call last):
  File ".\lczero-training-jio\tf\train.py", line 208, in <module>
    main(argparser.parse_args())
  File ".\lczero-training-jio\tf\train.py", line 148, in main
    workers=train_workers)
  File "C:\Users\Chad\Desktop\leela-training\lczero-training-jio\tf\chunkparser.py", line 128, in __init__
    p.start()
  File "C:\Users\Chad\AppData\Local\Programs\Python\Python37\lib\multiprocessing\process.py", line 112, in start
    self._popen = self._Popen(self)
  File "C:\Users\Chad\AppData\Local\Programs\Python\Python37\lib\multiprocessing\context.py", line 223, in _Popen
    return _default_context.get_context().Process._Popen(process_obj)
  File "C:\Users\Chad\AppData\Local\Programs\Python\Python37\lib\multiprocessing\context.py", line 322, in _Popen
    return Popen(process_obj)
  File "C:\Users\Chad\AppData\Local\Programs\Python\Python37\lib\multiprocessing\popen_spawn_win32.py", line 89, in __init__
    reduction.dump(process_obj, to_child)
  File "C:\Users\Chad\AppData\Local\Programs\Python\Python37\lib\multiprocessing\reduction.py", line 60, in dump
    ForkingPickler(file, protocol).dump(obj)
TypeError: can't pickle weakref objects
PS C:\Users\Chad\Desktop\leela-training> Traceback (most recent call last):
  File "<string>", line 1, in <module>
  File "C:\Users\Chad\AppData\Local\Programs\Python\Python37\lib\multiprocessing\spawn.py", line 105, in spawn_main
    exitcode = _main(fd)
  File "C:\Users\Chad\AppData\Local\Programs\Python\Python37\lib\multiprocessing\spawn.py", line 115, in _main
    self = reduction.pickle.load(from_parent)
EOFError: Ran out of input```

cn4750:
@Ghotrix wrote a PR to fix this in #81. Might be able to port those changes over.

Ttl:
I didn't know that there was an issue with Python 3.7. That PR probably applies with minimal conflicts to this PR too.

Tilps:
If we could fix the sampling dataset optimization in tf, maybe we could just use the experimental path and be done with it :P
I'd approve the virtual_batch_size thing in a separate PR in an instant - the rest needs a closer review than I've done so far.

cn4750:
Using an older Python, version 3.6, I've managed to get it working after using the shuffle.py script on 13 million games. This took 36 hours. Maybe this would be good functionality to add to the rescorer to help with speed?

These 100 game blocks now allows me to train off of HDD instead of SDD while adding about 20% more positions per second.

jjoshua2:
I've only been able to train 15.8k steps with this. Had 3 times now, it just hung there all around the same spot with no output and no gpu usage even all night, after 53458500 output and after 53458750 output. Seems very odd with random shuffling to be possible to hang up in almost the exact same spot but slightly different each time. Doesn't seem likely to be data corruption or hardware fault. It would seem to be something more like swa or momentum which wasn't changed...

jjoshua2:
I think it was something to do with bad displayport connection or graphics
driver not this code.

PR #105, opened 2020-02-26
Moves left head, tf2 version

Ttl:
Updated #84 for tf2.

Moves left head structure is changed from softmax to just predicting a scalar value. MSE loss didn't work with scalar, Huber works better. The head predicts plies left and expects the training data to also have plies left in the move_count field. Adding `plies_left = struct.pack('f', 2 * ply_count)` to chunkparser can be used to train from moves left data (As outputted by the current version of the rescorer.).

Also includes #104 because training without it uses too much memory.


Ttl:
Instructions for adding a moves left head to existing net:

In yaml add `moves_left: 'v1'` under `model`. Use `net_to_model.py` with flag `--ignore-errors` to replace weights with existing net without moves left head.

To train only the moves left head with other weights frozen add the following snippet in `init_v2` under mode creation on line 160:

```
for l in self.model.layers:
    if 'moves_left' not in l.name:
        l.trainable = False
```

ASilver:
I don't know if this code requires a new rescorer or rescoring, but the data that works fine with the first version is choking the new training code. I get as far as the typical message:

```
2020-02-26 15:38:50.591749: I tensorflow/stream_executor/platform/default/dso_loader.cc:44] Successfully opened dynamic library cudart64_101.dll
Using 1576 evaluation batches

```
And that is all she wrote. It never posts again, and a Ctrl-C shows it had issues with the actual games.

```
Process Process-46:
failed to parse C:\Users\Albert\Chess\NN_training\training\rescored2\game_016051.gz
failed to parse C:\Users\Albert\Chess\NN_training\training\rescored14\game_008630.gz
failed to parse C:\Users\Albert\Chess\NN_training\training\rescored7\game_022929.gz
failed to parse C:\Users\Albert\Chess\NN_training\training\rescored7\game_009406.gz
```

ASilver:
So the problem turned out to be the '_Faster less memory intensive input pipeline_' that was embedded. By excising it from the code the training no longer choked and began normally. This is with a Windows setup I need to add. There are a few details I will share with anyone wishing to try this on Windows themselves:

You will need the net.proto file from the first Moves Left PR. The current one will give you the wrong one if you go through "Git Clone...." and nothing will work. The second is an oddity if you do not build your own Tensorflow. I myself usually install with Pip. The page of requirements says 2.0 needs cuDNN 7.4 and CUDA 10.0, but neglects to mention that if you use 2.0.1 (a patched version of 2.0) then it now needs cuDNN 7.6 for CUDA 10.0. Easy to fix, but can be missed when it crashes and scrolls a mountain of messages.

tf/tfprocess.py
```diff
@@ -570,15 +687,20 @@ def calculate_swa_validations_v2(self, steps):
...
+        sum_moves_left = 0
+        sum_moves_left_mean_error = 0
         sum_mse = 0
         sum_policy = 0
         sum_value = 0
         counter = 0
-        for (x,y,z,q) in self.validation_dataset:
-            policy_loss, value_loss, mse_loss, policy_accuracy, value_accuracy = self.calculate_test_summaries_inner_loop(x, y, z, q)
+        for (x,y,z,q,m) in self.validation_dataset:
+            policy_loss, value_loss, moves_left_loss, mse_loss, policy_accuracy, value_accuracy = self.calculate_test_summaries_inner_loop(x, y, z, q, m)
```
Tilps: Should be assigning moves_left_mean_error here.

Ttl:
I trained some 128x10 networks from CCRL dataset for 100k steps using 2048 batch size with different moves left loss weights. Policy weight was 1.0 and value 0.5. Tensorboard stats:

ml_policy2_test

ml_value2_test

ml_moves_left

Adding moves left head didn't really change the policy loss, but did decrease the value loss.

Quick match with 2k fixed nodes, 5-piece TBs and moves left logic disabled to test only how the policy and value head strength was affected. Default parameters, but minibatch-size was set to 64:

```
   # PLAYER                      :  RATING  ERROR  POINTS  PLAYED   (%)  CFS(%)    W     D     L  D(%)
   1 lc0_ccrl_moves_left_1.0     :     9.8   15.9   428.0     800    54      57  244   368   188    46
   2 lc0_ccrl_moves_left_0.01    :     7.6   15.9   425.5     800    53      51  239   373   188    47
   3 lc0_ccrl_moves_left_0.1     :     7.2   15.8   425.0     800    53      77  224   402   174    50
   4 lc0_ccrl_moves_left_0.05    :    -2.1   16.4   414.5     800    52      65  226   377   197    47
   5 lc0_ccrl_moves_left_0.5     :    -7.5   16.1   408.5     800    51      79  233   351   216    44
   6 lc0_ccrl_no_moves_left      :   -15.0    6.5  1898.5    4000    47     ---  963  1871  1166    47
```

PR #106, opened 2020-03-04
Remove unnecessary tobytes calls

Ttl:
About 10% speedup on input pipeline limited cases.


Tilps:
Any chance there are python version specific issues that could be at play with this change?

Ttl:
I don't think so. This change just moves `tobytes` call to init instead of always calculating it again. It should work on all Python versions.

Tilps:
Ahh right, I misread the first line of the change.

PR #107, opened 2020-03-13
Don't block on full pipes

Ttl:
The current version can deadlock if a reader process dies and the pipe gets full. It shouldn't happen in normal operation but this is the only place I can think that could cause a deadlock. This commit checks if the chunk filename pipe is full and skips it in that case instead of blocking forever.

Works on Linux but needs testing on Windows before merging. @jjoshua2

About 1% slower than before on pipeline limited cases in my testing.

### DIFF
--- a/tf/chunkparser.py
+++ b/tf/chunkparser.py
@@ -26,6 +26,7 @@ import struct
 import tensorflow as tf
 import unittest
 import gzip
+from select import select
 
 V4_VERSION = struct.pack('i', 4)
 V3_VERSION = struct.pack('i', 3)
@@ -42,6 +43,9 @@ class ChunkDataSrc:
         return self.items.pop()
 
 
+def pipe_full(pipe):
+    r, w, x_ = select([], [pipe], [], 0)
+    return len(w) == 0
 
 def chunk_reader(chunk_filenames, output_pipes):
     """
@@ -60,9 +64,10 @@ def chunk_reader(chunk_filenames, output_pipes):
             print("chunk_reader didn't find any chunks.")
             return None
         while len(chunks):
-            filename = chunks.pop()
-            done.append(filename)
-            output_pipes[n].send(filename)
+            if not pipe_full(output_pipes[n]):
+                filename = chunks.pop()
+                done.append(filename)
+                output_pipes[n].send(filename)
             n += 1
             if n >= len(output_pipes):
                 n = 0


tf/chunkparser.py
```diff
@@ -60,9 +64,10 @@ def chunk_reader(chunk_filenames, output_pipes):
             print("chunk_reader didn't find any chunks.")
             return None
         while len(chunks):
-            filename = chunks.pop()
-            done.append(filename)
-            output_pipes[n].send(filename)
+            if not pipe_full(output_pipes[n]):
+                filename = chunks.pop()
+                done.append(filename)
+                output_pipes[n].send(filename)
```
Tilps: If one of the children is failed - its probably good to know.  How often will it false positive if we have a log message in else path?

jjoshua2:
Yes I think this was the cause of the issue I've been having. I did something similar to fix it but didnt notice any slowdown

jjoshua2:
Btw troubleshooting these things is so annoying. I had prints for each send and received of each pipe and lowered thread count down to 2.

tf/chunkparser.py
Ttl: I tested it and got multiple messages per second. Since reading filenames is so fast the pipes are almost always full.

Tilps: What do you think about registering an atexit to log that the subprocess is exiting?

Ttl: There should be exception printed to the console if a reader process crashes. It should also already print "Reader EOF" in `v4_gen` if a reader exits and the pipe is closed.

Tilps: That isn't what is happening at run2 training then.  Only standard log messages between training start and training stuck.

Tilps: Or child process I/O isn't being printed at all?

leedavid:
it will generate lot WARNING.  when run under windows 

WARNING:tensorflow:AutoGraph could not transform <bound method DescriptorBase.GetOptions of <google.protobuf.descriptor.Descriptor object at 0x0000019DEAB846D8>> and will run it as-is.
Please report this to the TensorFlow team. When filing the bug, set the verbosity to 10 (on Linux, `export AUTOGRAPH_VERBOSITY=10`) and attach the full output.
Cause: LIVE_VARS_IN
WARNING:tensorflow:AutoGraph could not transform <bound method DescriptorBase.GetOptions of <google.protobuf.descriptor.Descriptor object at 0x0000019DEAB846D8>> and will run it as-is.
Please report this to the TensorFlow team. When filing the bug, set the verbosity to 10 (on Linux, `export AUTOGRAPH_VERBOSITY=10`) and attach the full output.
Cause: LIVE_VARS_IN

leedavid:
Also this is a error in function pipe_full

Using 4 worker processes.
Process Process-5:
Traceback (most recent call last):
  File "C:\ProgramData\Anaconda3\envs\tf36\lib\multiprocessing\process.py", line 258, in _bootstrap
    self.run()
  File "C:\ProgramData\Anaconda3\envs\tf36\lib\multiprocessing\process.py", line 93, in run
    self._target(*self._args, **self._kwargs)
  File "D:\V100\TF20\tf\chunkparser.py", line 77, in chunk_reader
    if not pipe_full(output_pipes[n]):
  File "D:\V100\TF20\tf\chunkparser.py", line 47, in pipe_full
    _, w, _ = select([], [pipe], [], 0)
OSError: [WinError 10038] 在一个非套接字上尝试了一个操作。
Using 2 worker processes.

so I changed to 
def pipe_full(pipe):
    try:
        _, w, _ = select([], [pipe], [], 0)
        return len(w) == 0
    except:
        return False

can running in winodws.

PR #108, opened 2020-03-15
Queue for chunk filenames

Ttl:
Maybe this works on Windows? @jjoshua2 @leedavid


cn4750:
On Windows I never had issues with the old code, but this does seem to work fine too. It may be about 5% slower though.

PR #109, opened 2020-03-21
V5 training data support

Tilps:
Except for the FRC variant.

### DIFF
--- a/tf/chunkparser.py
+++ b/tf/chunkparser.py
@@ -206,7 +206,7 @@ class ChunkParser:
         # v3/4 data sometimes has a useful value in dep_ply_count, so copy that over if the new ply_count is not populated.
         if plies_left == 0:
             plies_left = dep_ply_count
-        m = struct.pack('f', plies_left)
+        plies_left = struct.pack('f', plies_left)
         # Don't yet support FRC input format decoding.
         assert input_format == 1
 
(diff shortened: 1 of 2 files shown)


tf/chunkparser.py
```diff
@@ -204,15 +192,23 @@ def convert_v4_to_tuple(self, content):
...
             float32 root_q (4 bytes)
             float32 best_q (4 bytes)
             float32 root_d (4 bytes)
             float32 best_d (4 bytes)
+            float32 root_m (4 bytes)
+            float32 best_m (4 bytes)
+            float32 plies_left (4 bytes)
         """
-        (ver, probs, planes, us_ooo, us_oo, them_ooo, them_oo, stm, rule50_count, ply_count, winner, root_q, best_q, root_d, best_d) = self.v4_struct.unpack(content)
+        (ver, input_format, probs, planes, us_ooo, us_oo, them_ooo, them_oo, stm, rule50_count, dep_ply_count, winner, root_q, best_q, root_d, best_d, root_m, best_m, ply_count) = self.v5_struct.unpack(content)
```
Ttl: `ply_count` could be renamed to `plies_left` as that's what it's referred to as elsewhere.

Tilps: plies_left already used for the output variable, so I would want to rename that one if I rename ply_count to plies_left.  But I'm not sure what to rename that...

Ttl: All other outputs are also named the same as variables unpacked from struct. I don't see a reason why it couldn't be `plies_left` everywhere.

Tilps: Yeah, okay, seeems inline with the rest of the function so, done.

PR #110, opened 2020-03-21
Faster way of doing experimental pipeline

Tilps:
Not quite as random, but probably good enough and I'm pretty sure a bit faster.

SKIP_MULTIPLE is a trade off of quality/speed for ram usage.

Still not as good as chunkparser though - I can only manage 3050pos/sec. Seems to decide to only use about 50% cpu load for some reason. Chunkparser is easier on the cpu, only 30% load and it gets 4250pos/sec

Might do a bit more fine tuning.

### DIFF
--- a/tf/train.py
+++ b/tf/train.py
@@ -29,6 +29,7 @@ from tfprocess import TFProcess
 from chunkparser import ChunkParser
 
 SKIP = 32
+SKIP_MULTIPLE = 1024
 
 
 def get_chunks(data_prefix):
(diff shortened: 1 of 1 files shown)


Tilps:
Now over 5500pos/sec on my machine. 70% gpu load, and the shuffle buffer fills at over 10k pos/sec

tf/train.py
```diff
@@ -158,8 +164,9 @@ def main(cmd):
 
     shuffle_size = int(shuffle_size*(1.0-train_ratio))
     if experimental_parser:
-        test_dataset = tf.data.Dataset.from_tensor_slices(test_chunks).shuffle(len(test_chunks)).repeat()\
-                         .interleave(lambda x: tf.data.FixedLengthRecordDataset(x, 8308, compression_type='GZIP', num_parallel_reads=1).filter(sample), num_parallel_calls=tf.data.experimental.AUTOTUNE)\
+        test_dataset = tf.data.Dataset.from_tensor_slices(test_chunks).shuffle(len(test_chunks)).repeat().batch(256)\
```
mooskagh: Why escape end of line? Leaving unmatched parentheses is more idiomatic way to do multiline python statements.
Also here `yapf` would also help. :-P

PR #111, opened 2020-03-22
Attempt to convert frc castling to input planes.

Tilps:
My brain hurts...

This needs lots of testing...


Tilps:
bitorder little introduces a dependency on numpy 1.17 - I might just rewrite it with a reversing slice.

Tilps:
I don't think I have a good way to test this yet - definitely some risk I've got the byte order or bit orders reversed in some place.

tf/chunkparser.py
```diff
@@ -207,20 +207,33 @@ def convert_v5_to_tuple(self, content):
...
         # detect edges of the board more easily
+        if input_format == 1:
+            castling_planes = self.flat_planes[us_ooo] + \
+                              self.flat_planes[us_oo] + \
+                              self.flat_planes[them_ooo] + \
+                              self.flat_planes[them_oo]
+        elif input_format == 2:
+            # Each inner array has to be reversed as these fields are in opposite endian to the planes data.
+            them_ooo_bytes = np.unpackbits(np.array([them_ooo], dtype=np.uint8))[:, ::-1].astype(np.float32).tobytes()
+            us_ooo_bytes = np.unpackbits(np.array([us_ooo], dtype=np.uint8))[:, ::-1].astype(np.float32).tobytes()
```
mooskagh:
```bash
pip install yapf
yapf --style=pep8 chunkparser.py  # or --style=google
```
would split those long lines. :-P

Tilps: Yes our training scripts could with a good formatting.

Tilps:
Step 1 to testing it is actually having training data with format 2, except that isn't step 1 because that requires a network with input format 2, except that isn't step 1 because that requires training code to make such a net.

So... I'm going to submit it for now under the assumption that I have at least properly guarded the new code away from the default case - and address any details once I write another PR to create FRC nets in the first place.

PR #116, opened 2020-03-25
Add input_type field to the model yaml section to label the saved net…

Tilps:
…works to produce alternative input format results.


Ttl:
This seems to just set the field in protobuf but there aren't any checks that the input data format in the training data matches. I think some checks would be good to have.

Tilps:
Done.

PR #118, opened 2020-03-27
clean up get_chunk functions

skondrashov:
1) remove get_chunks, since it's only used once and is a simple one-liner
2) remove copy/pasted code in get_latest_chunks
3) use f"" instead of .format


oscardssmith:
bump?

Tilps:
This still seems reasonable, but needs a conflict resolution.

PR #119, opened 2020-03-27
clean up extract_io in train.py

skondrashov:
1) use one statement per line where lines wrap from being too dense
2) use a function for repeated decode_raw(substr()) calls
3) more consistent comment style 
4) align identical sections of lines
5) ignore max columns to help readability

### DIFF
--- a/tf/train.py
+++ b/tf/train.py
@@ -69,44 +69,51 @@ def get_latest_chunks(path, num_chunks, allow_less):
 
 
 def extract_inputs_outputs(raw):
-    # first 4 bytes in each batch entry are boring.
+    def read(pos, len, type):
+        return tf.io.decode_raw(tf.strings.substr(raw, pos, len), type)
+
+    # First 4 bytes in each batch entry are boring.
+
     # Next 4 change how we construct some of the unit planes.
-    input_format = tf.reshape(
-        tf.io.decode_raw(tf.strings.substr(raw, 4, 4), tf.int32),
-        [-1, 1, 1, 1])
+    input_format = read(4, 4, tf.int32)
+    input_format = tf.reshape(input_format, [-1, 1, 1, 1])
 
     # Next 7432 are easy, policy extraction.
-    policy = tf.io.decode_raw(tf.strings.substr(raw, 8, 7432), tf.float32)
+    policy = read(8, 7432, tf.float32)
+
     # Next are 104 bit packed chess boards, they have to be expanded.
-    bit_planes = tf.expand_dims(
-        tf.reshape(
-            tf.io.decode_raw(tf.strings.substr(raw, 7440, 832), tf.uint8),
-            [-1, 104, 8]), -1)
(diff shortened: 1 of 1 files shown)


tf/train.py
```diff
@@ -69,44 +69,51 @@ def get_latest_chunks(path, num_chunks, allow_less):
...
-        1., tf.cast(bitsplat_unit_planes, tf.float32))
+    bit_planes = read(7440, 832, tf.uint8)
+    bit_planes = tf.reshape            (bit_planes, [-1, 104, 8])
+    bit_planes = tf.expand_dims        (bit_planes, -1)
+    bit_planes = tf.tile               (bit_planes, [1, 1, 1, 8])
+    bit_planes = tf.bitwise.bitwise_and(bit_planes, [128, 64, 32, 16, 8, 4, 2, 1])
+    bit_planes = tf.cast               (bit_planes, tf.float32)
+    bit_planes = tf.minimum            (bit_planes, 1.)
+
+    # Next 5 bytes are 1 or 0 to indicate 8x8 planes of 1 or 0.
```
tkondrashov: my best guess, not sure if this is what the original comment was saying

tf/train.py
```diff
@@ -69,44 +69,51 @@ def get_latest_chunks(path, num_chunks, allow_less):
...
-            tf.io.decode_raw(tf.strings.substr(raw, 8272, 5), tf.uint8), -1),
-        -1)
-    unit_planes = tf.tile(unit_planes, [1, 1, 8, 8])
-    # In order to do the conditional for frc we need to make bit unpacked versions.  Note little endian for these fields so the bitwise_and array is reversed.
-    bitsplat_unit_planes = tf.bitwise.bitwise_and(
-        unit_planes, [1, 2, 4, 8, 16, 32, 64, 128])
-    bitsplat_unit_planes = tf.minimum(
-        1., tf.cast(bitsplat_unit_planes, tf.float32))
+    bit_planes = read(7440, 832, tf.uint8)
+    bit_planes = tf.reshape            (bit_planes, [-1, 104, 8])
```
mooskagh: Note that we format code with `yapf`, which does [pep8](https://www.python.org/dev/peps/pep-0008/) compliant formatting, and this change seems to be a step out of that style. (see the last paragraph of https://www.python.org/dev/peps/pep-0008/#pet-peeves)

tkondrashov: yeah it does seem that way... I'll say I'm a huge proponent of using spaces to align repetitive blocks like this (you can see what it does for readability), but I'll change it to conform

tkondrashov:
addressed @mooskagh 's comment, fixed a bug that I somehow introduced after testing the first time too

Tilps:
I am thinking of abandoning the attempt to support multiple input formats in a single function and just have separate functions for each input format.  Which I'll do when I add input format 3.

Tilps:
Obsolete - the code has been deleted.

PR #120, opened 2020-03-27
clean up config section in train.py main

skondrashov:


tf/train.py
```diff
@@ -158,21 +158,20 @@ def main(cmd):
...
-                                         num_train, allow_less)
-        test_chunks = get_latest_chunks(cfg['dataset']['input_test'], num_test,
-                                        allow_less)
-    else:
-        chunks = get_latest_chunks(cfg['dataset']['input'], num_chunks,
-                                   allow_less)
-        if allow_less:
-            num_train = int(len(chunks) * train_ratio)
-            num_test = len(chunks) - num_train
+    if 'input' in cfg['dataset']:
```
tkondrashov: makes more sense to check for 'input' first than an arbitrary one of 'input_train' and 'input_test'

tf/train.py
```diff
@@ -158,21 +158,20 @@ def main(cmd):
...
+        path = cfg['dataset']['input']
```
tkondrashov: more readable way to deal with 80 column limit than splitting the line

tf/train.py
```diff
@@ -158,21 +158,20 @@ def main(cmd):
...
-    num_test = num_chunks - num_train
-    if 'input_test' in cfg['dataset']:
-        train_chunks = get_latest_chunks(cfg['dataset']['input_train'],
-                                         num_train, allow_less)
-        test_chunks = get_latest_chunks(cfg['dataset']['input_test'], num_test,
-                                        allow_less)
-    else:
-        chunks = get_latest_chunks(cfg['dataset']['input'], num_chunks,
-                                   allow_less)
-        if allow_less:
```
tkondrashov: we can slice even if allow_less is false, don't need a special case

PR #121, opened 2020-04-07
Add input type 3 support to training

Tilps:
Significant refactoring of experimental pipeline extractors to avoid nested tf.where based logic.


tf/tfprocess.py
```diff
@@ -133,6 +133,8 @@ def __init__(self, cfg):
             self.INPUT_MODE = pb.NetworkFormat.INPUT_CLASSICAL_112_PLANE
         elif input_mode == "frc_castling":
             self.INPUT_MODE = pb.NetworkFormat.INPUT_112_WITH_CASTLING_PLANE
+        elif input_mode == "cannonical":
```
Ttl, line 136: canonical

Tilps: Fixed, thanks.

PR #125, opened 2020-05-01
Bump minimum lc0 version when creating nets with new input types.

Tilps:


Dboingue:
any discussion crumbs about this?  on Discord.  Maybe some links would help.  or issues attached.
thanks for keeping documentation possible on a regular basis.  I know that if everyone, thinks a bit about audience not being exactly their own self, documentation might not be such a chose, to do after the mountain has risen....  this is meant to be encouraging.

PR #127, opened 2020-05-24
Needs protobuf in requirement.txt

TesseractA:
Adding protobuf 3.12.1 to the list of requirements for lczero-training (if I understand what I'm doing)

### DIFF
--- a/tf/requirements.txt
+++ b/tf/requirements.txt
@@ -1,3 +1,4 @@
 numpy==1.13.3
 tensorflow==2.0.1
 tensorflow-tensorboard==0.4.0rc2
+protobuf==3.12.1


mooskagh:
The correct way to make the correct requirements.txt (if you feel like fixing it:) :

10 Create a virtualenv so that no external packages are installed:
`$ virtualenv virtualenv`

20 Activate virtualenv:
`$ . virtualenv/bin/activate`

30 Try running the script. IF it succeeds THEN GOTO 60

40 (if it fails due to missing package):
`(virtualenv) $ pip install <missing package>`

50 GOTO 30

60 Run this:
```
(virtualenv) $ pip freeze > requirements.txt
```


That's it. :)
You can go directly to the line 60 and see what `pip freeze` would generate for you, but I expect it will contain hundreds of various packages not needed for the script.

PR #129, opened 2020-06-29
Add support for input format 4 and 132

Tilps:
Completely untested...


Tilps:
Tested both 4 and 132, seems fine.

PR #137, opened 2020-12-19
Add some extensions to chunk retrieval.

Tilps:
Specifically support for.
1) Sorting chunks to select newest by a method other than mtime, for systems where mtime retrieval is slow.
2) Supporting multiple input directory specification in case its not possible to glob the directory names.
3) Flush the sort messages so they show up before sorting completes.

### DIFF
--- a/tf/train.py
+++ b/tf/train.py
@@ -37,18 +37,25 @@ def get_chunks(data_prefix):
 
 
 def get_all_chunks(path):
+    if isinstance(path, list):
+        print("getting chunks for", path)
+        chunks = []
+        for i in path:
+            chunks += get_all_chunks(i)
+        return chunks
     chunks = []
     for d in glob.glob(path):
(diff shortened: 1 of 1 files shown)


brianprichardson:
It may not be an issue for RL, but for SL it is often the case that there are some subdirectories and many chunk files have the same file name (starting from game_000000.gz and contain multiple games from the trainingdata-tool), so could the sort also include the directory name in addition to the file name?

Maybe it does not really matter if they are all randomized in the shuffle buffer anyway.

Hope I'm not over-thinking this.
Thanks.

Tilps:
Name sort as implemented here includes the directory name.
Number sort does not - and I think that makes sense.
Ultimately the sort doesn't matter at all unless you have more games than your configured limit. We always immediately undo the sort by shuffling the chunk list as soon as its the right length. (Shuffler buffer shuffling is a separate topic.)

PR #138, opened 2021-01-10
Added a YAML option to disable most L2 regularization

Rocketknight1:
Hey, this is something I've been testing myself and seeing ELO gains with, and I wanted to create this PR to get some feedback. It works well for me, but I was testing in conjunction with a few other things (particularly Adam as an optimizer), so I'm not 100% sure that this change was the cause. However, I'm virtually certain it won't break anything, and I'm... let's say 99.5% sure it was a big part of the cause.

The underlying reasoning is simple: Most advice for training supervised models assumes the training data is limited compared to the model capacity - this is why tools like L2 regularization, dropout and so on are popular, as they alleviate overfitting. When training with Leela you have vast training data, and overfitting is essentially a non-issue. As such, at least in theory you shouldn't need any L2 regularization at all!

However, [as this blog post points out](https://blog.janestreet.com/l2-regularization-and-batch-norm/), L2 regularization actually fulfills a completely different role in networks with batchnorms, like Leela's - it has no regularization effect whatsoever, and instead prevents unwanted learning rate decay by keeping the norms of weight tensors from blowing up.

Thus, removing L2 regularization entirely would probably have some unwanted effects, by causing the effective LR to constantly decay in parts (but not all) of the network. However, we _can_ safely remove it for all weights and biases that do not precede batchnorms - these are the places where it actually has a regularizing effect, which is what we don't want.

Training with Adam and only regularizing pre-batchnorm weights gave me a bonus of about 20 ELO in a 128x10b network @ 1000 nodes per move. I suspect that Adam might have accelerated the training, but the change in regularization is almost certainly responsible for the better final accuracy. However, I understand if you want me to do some clear A/B testing to confirm that. Let me know what you think, or if you'd prefer I bundle all of the changes I've been trying (Adam + fixed Adam regularization + only prenorm regularization) into one big PR.

### DIFF
--- a/tf/tfprocess.py
+++ b/tf/tfprocess.py
@@ -70,6 +70,7 @@ class TFProcess:
         self.RESIDUAL_BLOCKS = self.cfg['model']['residual_blocks']
         self.SE_ratio = self.cfg['model']['se_ratio']
         self.policy_channels = self.cfg['model'].get('policy_channels', 32)
+        self.only_prenorm_regularization = self.cfg['model'].get('only_prenorm_regularization', False)
         precision = self.cfg['training'].get('precision', 'single')
         loss_scale = self.cfg['training'].get('loss_scale', 128)
         self.virtual_batch_size = self.cfg['model'].get(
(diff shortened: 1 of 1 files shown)


Rocketknight1:
Closing this PR for now - after several additional runs, I don't think removing L2 regularization from the output heads was the reason for the reduced training loss in my runs. 

The reduced training loss is a very real phenomenon though, and I'm going to do ablation testing to figure out which thing I changed is the cause. There'll probably be another PR then :p

PR #139, opened 2021-01-24
V6 training format and value focus

jhorthos:
Implement V6 training format and a first version of value focus in training.

### DIFF
--- a/tf/chunkparser.py
+++ b/tf/chunkparser.py
@@ -33,8 +33,9 @@ V5_VERSION = struct.pack('i', 5)
 CLASSICAL_INPUT = struct.pack('i', 1)
 V4_VERSION = struct.pack('i', 4)
 V3_VERSION = struct.pack('i', 3)
-# jhorthos query - not sure i understand Struct format strings fully
-V6_STRUCT_STRING = '4si7432s832sBBBBBBBbfffffffffffffffIHH4H'
+# jhorthos query - not sure i understand Struct format strings fully B? b?
+# followed by 15 floats, 1 uint_32, 2 uint_16, and 4x16 reserved
+V6_STRUCT_STRING = '4si7432s832sBBBBBBBbBfffffffffffffffIHH4H'
 V5_STRUCT_STRING = '4si7432s832sBBBBBBBbfffffff'
 V4_STRUCT_STRING = '4s7432s832sBBBBBBBbffff'
 V3_STRUCT_STRING = '4s7432s832sBBBBBBBb'
@@ -99,6 +100,7 @@ class ChunkParser:
         'chunks' list of chunk filenames.
         'shuffle_size' is the size of the shuffle buffer.
         'sample' is the rate to down-sample.
+        'value_focus_min' and 'value_focus_slope' control value focus
         'workers' is the number of child workers to use.
 
         The data is represented in a number of formats through this dataflow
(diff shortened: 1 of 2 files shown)


tf/chunkparser.py
```diff
@@ -28,10 +69,14 @@
...
 from select import select
 
+V6_VERSION = struct.pack('i', 6)
 V5_VERSION = struct.pack('i', 5)
 CLASSICAL_INPUT = struct.pack('i', 1)
 V4_VERSION = struct.pack('i', 4)
 V3_VERSION = struct.pack('i', 3)
+# jhorthos query - not sure i understand Struct format strings fully B? b?
+# followed by 15 floats, 1 uint_32, 2 uint_16, and 64 (4x16) reserved
+V6_STRUCT_STRING = '4si7432s832sBBBBBBBBfffffffffffffffIHH4H'
```
Tilps, line 79: While Borg changed v5 result from signed to unsigned dummy in v6, I'm not sure we should respect that here - since it would allow for simplification of convert_v6_to_tuple. (And dummy is unused so interpreting it as per the v6  spec has no actual effect)

tf/chunkparser.py
```diff
@@ -193,10 +245,65 @@ def parse_function(planes, probs, winner, q, plies_left):
...
+        uint32_t visits;                             8340
+        // Indices in the probabilities array.
+        uint16_t played_idx;                         8344
+        uint16_t best_idx;                           8346
+        uint64_t reserved;                           8348
         """
-        Unpack a v5 binary record to 5-tuple (state, policy pi, result, q, m)
 
+        version = content[0:4]
+        if version == V6_VERSION:
```
Tilps, line 299: If you leave dummy as a signed value - we can skip this conditional and just unconditionally unpack as v6_struct.

tf/chunkparser.py
```diff
@@ -218,14 +325,17 @@ def convert_v5_to_tuple(self, content):
...
-        if plies_left == 0:
-            plies_left = dep_ply_count
-        plies_left = struct.pack('f', plies_left)
 
+        # jhorthos query - this is the same as V5 code - still valid?
+        else:
+            (ver, input_format, probs, planes, us_ooo, us_oo, them_ooo, them_oo,
+             stm, rule50_count, dep_ply_count, winner, root_q, best_q, root_d,
+             best_d, root_m, best_m, plies_left) = self.v5_struct.unpack(content)
+            # v3/4 data sometimes has a useful value in dep_ply_count, so copy that over if the new ply_count is not populated.
```
Tilps, line 334: If you follow my suggestion regarding always unpacking with v6 struct - keep this code to do with plies_left.  It'll never fire for proper v5 or v6 data, only for v3/4.

tf/chunkparser.py
```diff
@@ -278,24 +388,36 @@ def convert_v5_to_tuple(self, content):
                  self.flat_planes[1]
 
         assert len(planes) == ((8 * 13 * 1 + 8 * 1 * 1) * 8 * 8 * 4)
-        winner = float(winner)
-        assert winner == 1.0 or winner == -1.0 or winner == 0.0
-        winner = struct.pack('fff', winner == 1.0, winner == 0.0,
+
+        # jhorthos query - pls check this
+        if version == V6_VERSION:
```
Tilps, line 393: This could be `if ver == 6:` to avoid having to extract 'version' at all if you follow the suggestion above.

tf/chunkparser.py
```diff
@@ -278,24 +388,36 @@ def convert_v5_to_tuple(self, content):
...
+            winner = struct.pack('fff', (1.0 - result_d + result_q)/2, result_d,
```
Tilps, line 394: For consistency with below - rather than `/2` maybe `* 0.5`

tf/chunkparser.py
```diff
@@ -278,24 +388,36 @@ def convert_v5_to_tuple(self, content):
...
+            winner = struct.pack('fff', winner == 1.0, winner == 0.0,
                              winner == -1.0)
+                          
 
         best_q_w = 0.5 * (1.0 - best_d + best_q)
         best_q_l = 0.5 * (1.0 - best_d - best_q)
         assert -1.0 <= best_q <= 1.0 and 0.0 <= best_d <= 1.0
         best_q = struct.pack('fff', best_q_w, best_d, best_q_l)
-
+
```
Tilps, line 407: Looks like some stray white space - should probably run your code change through yapf.

tf/chunkparser.py
```diff
@@ -309,7 +431,20 @@ def sample_record(self, chunkdata):
                 # Downsample, using only 1/Nth of the items.
                 if random.randint(0, self.sample - 1) != 0:
                     continue  # Skip this record.
+                    
             record = chunkdata[i:i + record_size]
+            
+            # peak at best_q and orig_q from record
+            best_q = struct.unpack('f', record[8284:8288])
+            orig_q = struct.unpack('f', record[8328:8332])
```
Tilps, line 439: This code is unsafe here - when upgrading you are attempting to read off the end of the record.  You either need to conditional this logic based on version == V6_VERSION: or move it after appending the additional bytes - or both.

tf/chunkparser.py
```diff
@@ -278,24 +388,36 @@ def convert_v5_to_tuple(self, content):
                  self.flat_planes[1]
```
Tilps, line 388: A few lines above here dep_ply_count is used - your v6 unpack line unpacks that into a variable called invariance_info you should rename as appropriate.

tf/chunkparser.py
Tilps: note that dep_ply_count is called invariance_info in your v6 unpack. So you'll need to rename.

tf/chunkparser.py
```diff
@@ -278,24 +388,36 @@ def convert_v5_to_tuple(self, content):
...
-        winner = struct.pack('fff', winner == 1.0, winner == 0.0,
+
+        # jhorthos query - pls check this
+        if version == V6_VERSION:
+            winner = struct.pack('fff', (1.0 - result_d + result_q)/2, result_d,
+                              (1.0 - result_d - result_q)/2)
+        else:
+            winner = float(winner)
+            assert winner == 1.0 or winner == -1.0 or winner == 0.0
+            winner = struct.pack('fff', winner == 1.0, winner == 0.0,
```
Tilps, line 399: if you do the uncoditional unpack you should either rename 'dummy' to 'winner' - or  rename all these 'winner' to 'dummy'. (or some other combination - like renaming both to dep_result.)

tf/chunkparser.py
jhorthos: okay i get all this up to here. mostly struggling to figure out git now. :-)

tf/chunkparser.py
jhorthos: ah yes

tf/chunkparser.py
jhorthos: fixed

tf/chunkparser.py
jhorthos: seems most sensible to leave unpack name 'dummy' doesn't it?

tf/chunkparser.py
jhorthos: oops

tf/chunkparser.py
jhorthos: fixed

tf/chunkparser.py
jhorthos: fixed, unconditional V6 unpack now

tf/chunkparser.py
jhorthos: ok

tf/chunkparser.py
jhorthos: should probably be rechecked - i think this is fixed now except the dummy/winner issue

tf/chunkparser.py
```diff
@@ -87,21 +130,24 @@ def __init__(self,
...
+        'value_focus_min' and 'value_focus_slope' control value focus
         'workers' is the number of child workers to use.
 
         The data is represented in a number of formats through this dataflow
         pipeline. In order, they are:
 
         chunk: The name of a file containing chunkdata
 
-        chunkdata: type Bytes. Multiple records of v5 format where each record
+        chunkdata: type Bytes. Multiple records of v5 or v6 format where each record
```
Tilps, line 150: Not sure if this comment is being clear by listing both v5 and v6.

tf/chunkparser.py
Tilps: `        if (input_format == 132 or input_format == 133) and dep_ply_count >= 128:`

tf/chunkparser.py
Tilps: The conditional is still here.

tf/chunkparser.py
Tilps: I like dep_result myself - but dummy is fine.  Either way this needs updating.

tf/chunkparser.py
```diff
@@ -278,24 +380,35 @@ def convert_v5_to_tuple(self, content):
...
 
         assert len(planes) == ((8 * 13 * 1 + 8 * 1 * 1) * 8 * 8 * 4)
-        winner = float(winner)
-        assert winner == 1.0 or winner == -1.0 or winner == 0.0
-        winner = struct.pack('fff', winner == 1.0, winner == 0.0,
+
+        # jhorthos query - pls check this
+        if ver == V6_VERSION:
+            winner = struct.pack('fff', 0.5 * (1.0 - result_d + result_q), result_d,
+                              (1.0 - result_d - result_q)/2)
```
Tilps, line 387: both changed to 0.5 * please.

tf/chunkparser.py
Tilps: Still seeing stray white space in this file.

tf/train.py
```diff
@@ -408,6 +408,9 @@ def main(cmd):
     # Load data with split batch size, which will be combined to the total batch size in tfprocess.
     ChunkParser.BATCH_SIZE = split_batch_size
 
+    value_focus_min = cfg['training'].get('value_focus_min', 1)
```
Tilps, line 411: Probably should throw an exception if these are non default values and experimental parser is enabled.

tf/chunkparser.py
jhorthos: ugh

Tilps (commented):
Seems fine to my eyes. True test is at runtime ;)

jhorthos:
will probably get a chance to test this weekend at the lastest. :-)

tf/chunkparser.py
```diff
@@ -278,10 +379,16 @@ def convert_v5_to_tuple(self, content):
...
-                             winner == -1.0)
+
+        # jhorthos query - pls check this
+        if ver == V6_VERSION:
+            winner = struct.pack('fff', 0.5 * (1.0 - result_d + result_q), result_d,
+                              0.5 * (1.0 - result_d - result_q))
+        else:
+            winner = float(dep_result)
+            assert dep_result == 1.0 or dep_result == -1.0 or dep_result == 0.0
+            dep_result = struct.pack('fff', dep_result == 1.0, dep_result == 0.0,
```
Tilps, line 390: On this line the first dep_result should be winner - since its what is returned.

tf/chunkparser.py
```diff
@@ -278,10 +379,16 @@ def convert_v5_to_tuple(self, content):
...
-        assert winner == 1.0 or winner == -1.0 or winner == 0.0
-        winner = struct.pack('fff', winner == 1.0, winner == 0.0,
-                             winner == -1.0)
+
+        # jhorthos query - pls check this
+        if ver == V6_VERSION:
+            winner = struct.pack('fff', 0.5 * (1.0 - result_d + result_q), result_d,
+                              0.5 * (1.0 - result_d - result_q))
+        else:
+            winner = float(dep_result)
```
Tilps, line 388: winner on this line should be dep_result (unlike below)

jhorthos: aha. good.
i will add the exception in train.py before long too.

tf/train.py
```diff
@@ -418,6 +418,10 @@ def main(cmd):
     experimental_reads = max(2, mp.cpu_count() - 2) // 2
     extractor = select_extractor(tfprocess.INPUT_MODE)
 
+    if experimental_parser and (not value_focus_min == 1 or not value_focus_slope == 0):
+            raise ValueError('Experimental parser does not support non-default value \
```
Tilps, line 422: This formatting looks a bit strange.
Also using not blah == 1 instead of blah != 1 feels wrong to me, but I'm not a python expert :P

Tilps: Still looks like excessive indentation on the next line.

jhorthos: oh okay i guess

Tilps (approved):

PR #141, opened 2021-02-24
Enable value focus in test set as well as train.

jhorthos:


tf/train.py
```diff
@@ -465,6 +465,8 @@ def read(x):
                                   shuffle_size=shuffle_size,
                                   sample=SKIP,
                                   batch_size=ChunkParser.BATCH_SIZE,
+                                  value_focus_min=value_focus_min,
```
Tilps, line 468: The comment above this is no longer correct.

Tilps:
This is obsolete because of code changes, and also I don't think we want it.
What we might want is a 3rd entirely line, 'biased test'.  However that is a much different PR, so I am going to close this one.

PR #143, opened 2021-03-12
distributed training on linux

masterkni666:
Never really checked if training produces correct values, leaving as draft. Created from Tilps' distribute branch
Need to set gpu field to 'all' to actually use it


teck45:
we need to test this same training data and then compare the networks. Who has GPU's? :)

tf/train.py
```diff
@@ -471,7 +470,9 @@ def read(x):
             output_types=(tf.string, tf.string, tf.string, tf.string,
                           tf.string))
         test_dataset = test_dataset.map(ChunkParser.parse_function)
-        test_dataset = test_dataset.prefetch(4)
+        if tfprocess.strategy is None: #Mirrored strategy appends prefetch itself with a value depending on number of replicas
```
Tilps, line 473: I think this is indented too much - it should run regardless of experimental_parser being true or false...

Tilps:
Looks like it does nothing unless 'all' is defined - which is good. I think it would be fine to commit and then refine under testing if there are issues found.

tf/train.py
masterkni666: fixed

tf/tfprocess.py
```diff
@@ -172,14 +177,28 @@ def __init__(self, cfg):
...
+        if self.strategy is not None:
+            self.test_dataset = self.strategy.experimental_distribute_dataset(test_dataset)
+        else:
+            self.test_dataset = train_dataset
+        self.test_iter = iter(self.test_dataset)
+        if self.strategy is not None and validation_dataset is not None:
+            self.validation_dataset = self.strategy.experimental_distribute_dataset(validation_dataset)
+        else:
+            self.validation_dataset = validation_dataset
+        this = self
```
Tilps, line 195: Curious trick - I think you can move it inside the if block though?

masterkni666: some java thing I picked up some time ago :p Anyways moved.

tf/tfprocess.py
```diff
@@ -191,9 +191,9 @@ def init_v2(self, train_dataset, test_dataset, validation_dataset=None):
         if self.strategy is not None and validation_dataset is not None:
             self.validation_dataset = self.strategy.experimental_distribute_dataset(validation_dataset)
         else:
-            self.validation_dataset = validation_dataset
-        this = self
+            self.validation_dataset = validation_dataset       
         if self.strategy is not None:
+        this = self
```
Tilps, line 196: you forgot the indent - for the python!

Can you run the entire file through yapf? After you fix that?

masterkni666: yike stupid mistake. I ran it through yapf, not sure if you wanted a particular style, I left it at default

Tilps (approved):

Tilps:
Default is fine IIRC.

tf/tfprocess.py
```diff
@@ -741,8 +858,12 @@ def calculate_test_summaries_v2(self, test_batches, steps):
...
             x, y, z, q, m = next(self.test_iter)
-            policy_loss, value_loss, moves_left_loss, mse_loss, policy_accuracy, value_accuracy, moves_left_mean_error, policy_entropy, policy_ul = self.calculate_test_summaries_inner_loop(
-                x, y, z, q, m)
+            if self.strategy is not None:
+                policy_loss, value_loss, moves_left_loss, mse_loss, policy_accuracy, value_accuracy, moves_left_mean_error, policy_entropy, policy_ul = self.strategy_calculate_test_summaries_inner_loop(
+                    x, y, z, q, m)
+            else:
+                policy_loss, value_loss, moves_left_loss, mse_loss, policy_accuracy, value_accuracy, moves_left_mean_error, policy_entropy, policy_ul = self.calculate_test_summaries_inner_loop(
+                    x, y, z, q, m)
             sum_policy_accuracy += policy_accuracy
```
Tilps, line 867: This block of delta - needs to be replicated added under the calculate validation code further down.  Validations were added since my original PR - and I guess you don't have them configured.

tf/train.py
```diff
@@ -471,6 +475,8 @@ def read(x):
             output_types=(tf.string, tf.string, tf.string, tf.string,
                           tf.string))
         test_dataset = test_dataset.map(ChunkParser.parse_function)
+    if tfprocess.strategy is None:  #Mirrored strategy appends prefetch itself with a value depending on number of replicas
+        train_dataset = train_dataset.prefetch(4)
```
Tilps, line 479: I think you'll need to do the same for the validation data set down below.

tf/train.py
```diff
@@ -475,16 +475,15 @@ def read(x):
...
     if 'input_validation' in cfg['dataset']:
         valid_chunks = get_all_chunks(cfg['dataset']['input_validation'])
         validation_dataset = tf.data.FixedLengthRecordDataset(valid_chunks, 8308, compression_type='GZIP', num_parallel_reads=experimental_reads)\
-                               .batch(split_batch_size, drop_remainder=True).map(extractor).prefetch(4)
-
+                               .batch(split_batch_size, drop_remainder=True).map(extractor)
+    if tfprocess.strategy is None:  #Mirrored strategy appends prefetch itself with a value depending on number of replicas
+        train_dataset = train_dataset.prefetch(4)
+        test_dataset = test_dataset.prefetch(4)
+        validation_dataset = validation_dataset.prefetch(4)
```
Tilps, line 486: This line needs to check validation_dataset is not None maybe?

Tilps (approved):

PR #145, opened 2021-03-14
Gpu selection

OfekShochat:
can select gpus for training. syntax is 
```
gpu: 1, 2, ..., n
```
and 
```
gpu: 1,2,...,n
```

### DIFF
--- a/tf/tfprocess.py
+++ b/tf/tfprocess.py
@@ -160,10 +160,10 @@ class TFProcess:
         self.renorm_momentum = self.cfg['training'].get(
             'renorm_momentum', 0.99)
    
-        if self.cfg['gpu'] == 'all':
+        if str(self.cfg['gpu']) == 'all':
             self.strategy = tf.distribute.MirroredStrategy()
             tf.distribute.experimental_set_strategy(self.strategy)
-        elif "," in self.cfg['gpu']:
+        elif "," in str(self.cfg['gpu']):
             active_gpus = []
             gpus = tf.config.experimental.list_physical_devices('GPU')
             for i in self.cfg['gpu'].split(","):
--- a/tf/tfprocess.py
+++ b/tf/tfprocess.py
@@ -159,10 +159,15 @@ class TFProcess:
         self.renorm_max_d = self.cfg['training'].get('renorm_max_d', 0)
         self.renorm_momentum = self.cfg['training'].get(
             'renorm_momentum', 0.99)
-
+   
         if self.cfg['gpu'] == 'all':
             self.strategy = tf.distribute.MirroredStrategy()
             tf.distribute.experimental_set_strategy(self.strategy)
+        elif self.cfg['gpu'].__contains__(','):
+            gpus = tf.config.experimental.list_physical_devices('GPU')
+            for i in self.cfg['gpu'].split(",")
+                tf.config.experimental.set_visible_devices(gpus[int(i)],
+                                                           'GPU')
         else:
             gpus = tf.config.experimental.list_physical_devices('GPU')
             print(gpus)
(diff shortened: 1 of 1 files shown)


tf/tfprocess.py
```diff
@@ -159,10 +159,15 @@ def __init__(self, cfg):
         self.renorm_max_d = self.cfg['training'].get('renorm_max_d', 0)
         self.renorm_momentum = self.cfg['training'].get(
             'renorm_momentum', 0.99)
-
+   
         if self.cfg['gpu'] == 'all':
             self.strategy = tf.distribute.MirroredStrategy()
             tf.distribute.experimental_set_strategy(self.strategy)
+        elif self.cfg['gpu'].__contains__(','):
```
Tilps, line 166: the more python way is `',' in self.cfg['gpu']` I think

tf/tfprocess.py
```diff
@@ -159,10 +159,15 @@ def __init__(self, cfg):
...
+            gpus = tf.config.experimental.list_physical_devices('GPU')
+            for i in self.cfg['gpu'].split(","):
+                tf.config.experimental.set_visible_devices(gpus[int(i)],
```
Tilps, line 169: This doesn't give you multiple gpus active - you need strategy for that.  What you want to do I think is make a list and add gpus[int(i)] for each i.  Then pass that to tf.distribute.MirroredStrategy initializer after the loop - then rest of the code is same as the 'all' case.

Tilps (dismissed):

tf/tfprocess.py
OfekShochat: ok sure

OfekShochat: ohh i thought I added that
sorry didn't see

OfekShochat:
@Tilps how can I push from gpu-selection branch on the fork to here now? I made the changes

Tilps:
nothing to do.

tf/tfprocess.py
```diff
@@ -159,10 +159,17 @@ def __init__(self, cfg):
         self.renorm_max_d = self.cfg['training'].get('renorm_max_d', 0)
         self.renorm_momentum = self.cfg['training'].get(
             'renorm_momentum', 0.99)
-
+
```
Tilps, line 162: some stray whitespace added here.

Tilps:
Needs a test and formatting fix and then probably fine to go.

tf/tfprocess.py
OfekShochat: k, so this is ok? no further changes?

dje-dev:
This enhancement will be very useful. However it seems to break backwards compatibility, for example a value of 0 now throws
  TypeError: argument of type 'int' is not iterable
instead perhaps the isinstance clause could be added as follows:
  elif isinstance(self.cfg['gpu'], str) and "," in self.cfg['gpu']:

OfekShochat:
yea yea its known. just I didnt know if this is even something to work on. if you are interested sure ill fix it happily!

OfekShochat:
well now its saying some git shit, but this should work now. can you check @dje-dev? not home...

dje-dev:
It turns out MirroredStrategy wants a set of string such as ["GPU:0", "GPU:1"].

So I was able to get code as shown below to work.

Note that set_memory_growth is called here, I'm not sure if that is required/desirable.
If it is called, it seems necessary to call it for all GPUs not just the selected ones, otherwise Tensorflow throws an error.
```
        elif "," in str(self.cfg['gpu']):
            gpus = tf.config.experimental.list_physical_devices('GPU')
            for gpu in gpus:
                tf.config.experimental.set_memory_growth(gpu, True)
            for i in self.cfg['gpu'].split(","):
                active_gpus.append("GPU:" + i)
            self.strategy = tf.distribute.MirroredStrategy(active_gpus)
            tf.distribute.experimental_set_strategy(self.strategy)
```

OfekShochat:
Kk
Works now? If it does I'll change my pr

PR #147, opened 2021-03-15
Multigpu startup performance improvement

Tilps:
The grads were unused, so forcing autograph to try and setup everything to compute each individual mean across the replicas was pointless and very expensive.
Saves 80 seconds for first train step on RAF's training machine.


kiudee (approved):
Looks good to me.

Was `grads` used at some point or why did we do that calculation?

Tilps:
Grads I believe we used to include in tensorboard - but we dropped them as part of rationalizing our tensorboard data.

PR #150, opened 2021-03-28
Policy focus

jhorthos:
implements policy focus in a form identical on the training side to value focus. requires merger of lc0 Policy focus #1535 to be used but shouldn't break anything until then.

also adds an option to use value focus (and policy focus) on the training test set if desired.

for both cases, the values for min and slope will be used for both train and test (if specified) datasets.

Settings for value focus and policy focus are set in the training yaml as follows:

```
training:
   value_focus_min: 0.6        # float, defaults to 1, range [0,1], 1 means value focus will not be active
   value_focus_slope: 0.5      # float, defaults to 0, range [0,+inf]
   value_focus_in_test: true   # defaults to false

   policy_focus_min: 0.6        # float, defaults to 1, range [0,1], 1 means policy focus will not be active
   policy_focus_slope: 0.5      # float, defaults to 0, range [0,+inf]
   policy_focus_in_test: true   # defaults to false  
```

### DIFF
--- a/tf/train.py
+++ b/tf/train.py
@@ -410,7 +410,22 @@ def main(cmd):
 
     value_focus_min = cfg['training'].get('value_focus_min', 1)
     value_focus_slope = cfg['training'].get('value_focus_slope', 0)
+    # defaults to not use value focus in test data
+    vfit = cfg['training'].get('value_focus_in_test', False)
+    if not vfit:
+        vf_test_min = 1  # min=1 inactivates value focus
+    else:
+        vf_test_min = value_focus_min
 
+    policy_focus_min = cfg['training'].get('policy_focus_min', 1)
+    policy_focus_slope = cfg['training'].get('policy_focus_slope', 0)
+    # defaults to not use policy focus in test data
+    pfit = cfg['training'].get('policy_focus_in_test', False)
+    if not pfit:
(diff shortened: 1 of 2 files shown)


tf/train.py
```diff
@@ -459,12 +476,15 @@ def read(x):
...
     else:
-        # no value focus for test_parser
         test_parser = ChunkParser(test_chunks,
                                   tfprocess.INPUT_MODE,
                                   shuffle_size=shuffle_size,
                                   sample=SKIP,
                                   batch_size=ChunkParser.BATCH_SIZE,
+                                  value_focus_min=vf_test_min,
+                                  value_focus_slope=value_focus_slope,
+                                  policy_focus_min=policy_focus_min,
```
Tilps, line 486: pf_test_min ?

Tilps:
I wonder what a 'useful' value for policy slope will end up being. I kind of suspect it'll be quite different to value.

kiudee (approved):

Tilps:
On training machine we're using a combined sampling function rather than two separate sampling functions. I'll probably look to create an alternative PR using that approach.

Tilps:
I've brought over 'diff focus' during my periodic resync of training code and the github and other cleanups I did today. So this is obsolete.

PR #162, opened 2021-07-28
Correctly calculate gradient clipping on aggregate gradients in multi-gpu

Tilps:
Before this change, each gpu clipped to its own norm - which could mean that its contribution to the total gradient got scaled unevenly.
This would then result in the applied gradient being in the wrong direction compared to both parts being scaled equally.

NOTE: This is a hack.  gradient_aggregator on Optimizer is not a documented public property.  It is especially undocumented that it contains the default aggregator value if None was passed at construction.  We need this default aggregator, because the 'documented' function to call to aggregate things crashes out with our grads as input. 

I suspect it might have something to do with None grads filtering that the default aggregator does (since None grads aren't variables, they don't have replica context attached to them or something and thus can't be reduced.) But I didn't dig deep enough to confirm that was the problem.

Another alternative approach which wouldn't be relying on undocumented behavior would be to do the clipping in the optimizer path itself - but it is very much unclear whether we could still graph grad_norm at all if we did that... I suspect not.

### DIFF
--- a/tf/tfprocess.py
+++ b/tf/tfprocess.py
@@ -592,13 +592,18 @@ class TFProcess:
         return policy_loss, value_loss, mse_loss, moves_left_loss, reg_term, new_grads
 
     def apply_grads(self, grads, effective_batch_splits):
+        grads = [
+            g[0] for g in self.orig_optimizer.gradient_aggregator(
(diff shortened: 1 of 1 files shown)


Tilps:
Been using this on training machine for a while without issue.

PR #164, opened 2021-08-10
New yaml-parameter: Value focus max

hans-ekbrand:
Cap the probability of inclusion into training for a position, regardless of its q.diff.

This idea was used to create the net posted here: https://discord.com/channels/425419482568196106/539960268982059008/874663457830633542

### DIFF
--- a/tf/chunkparser.py
+++ b/tf/chunkparser.py
@@ -169,6 +169,7 @@ class ChunkParser:
         self.sample = sample
         # set the min and slope for value focus, defaults accept all positions
         self.value_focus_min = value_focus_min
+        self.value_focus_max = value_focus_max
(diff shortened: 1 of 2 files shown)


hans-ekbrand:
The code is working, but I haven't tested the performance properly yet.

Tilps:
I think mathematically - this should be equivalent to increasing the slope and the min.  Specifically a 0.3 max is equivalent to multiplying slope and min by 3.3. (To get very exactly equivalent also scale SKIP by 3.3 - but scaling SKIP probably isn't needed to be 'good enough equivalence'.)

hans-ekbrand:
Good point, but it is still convenient to have this parameter, especially since SKIP is not a yaml-parameter.

PR #178, opened 2021-10-31
Output plies_left with the correct dimensions.

Tilps:


Tilps:
This was causing moves left mean error to be calculated incorrectly.
Unclear if it was also breaking the Huber Loss used for moves left loss calculation itself... The losses don't seem to move much...

PR #188, opened 2021-12-29
Allow to choose which blocks have SE units

Tilps:
Inspired by testing which suggests a 15 block net only needs 3 SE units to get the same loss values.


Tilps:
yaml options.
No se_blocks section means same as before, all blocks have SE.
`se_blocks: []` means no SE at all.
```
se_blocks:
  - 1 based block numbers
  - listed here
  - to control exactly which
  ```

PR #200, opened 2022-04-28
update init.sh to print useful info

teck45:
init.sh will print ls command of folder with proto files, so user will see they exist and successfully patched

### DIFF
--- a/init.sh
+++ b/init.sh
@@ -3,3 +3,6 @@
 protoc --proto_path=libs/lczero-common --python_out=tf libs/lczero-common/proto/net.proto
 protoc --proto_path=libs/lczero-common --python_out=tf libs/lczero-common/proto/chunk.proto
 touch tf/proto/__init__.py
+echo "cd libs/lczero-common/proto && ls"
+cd libs/lczero-common/proto && ls
+echo proto files are successfully patched


mooskagh:
Doesn't seem very useful to be honest.

teck45:
It is useful, because if person sees some output he/she will remember it was done, also print will tell what actually was done and where are those files are, useful in case they are missing.

teck45:
For example contributing client don't need to output thousands of lines - one per each generated game. That's why we limit it with on collar scripts
```
... 2>&1 >/dev/null | grep --line-buffered -v trainDir | sed --unbuffered -n '0~{GAMES}p'
```
But zero output is also wrong, especially when operation is done once by person looking at monitor. If we renaming something we print what was done, moving - also from and where, and so on. Computer should communicate =)
PS:
this is what bash script launching 4 clients shows:
```
admin@a4000-21bn12:~$ ./start.sh
   __        /\_/\                         
  / /  _____= o_o =                        
 ( (_./  )_    ^__                         
  \__(____)__<___>(@)  CLIENTS STARTED   
```

teck45:
I'm not saying print statement there is life changing piece of code, but it improves usability =)

PR #211, opened 2022-12-13
Attention net body protobuf converters for smolgen weights

almaudoh:
Tensorflow to Protobuf converter for smolgen weights


almaudoh:
Changes have been added in PR #212

PR #212, opened 2023-03-17
Training transformers smolgen

masterkni6:
generates trained T80 and smolgen transformer models into nets fine


tf/tfprocess.py
```diff
@@ -113,16 +141,42 @@ def __init__(self, cfg):
...
+            self.embedding_size = self.RESIDUAL_FILTERS
+            
         self.policy_channels = self.cfg['model'].get('policy_channels', 32)
         self.pol_embedding_size = self.cfg['model'].get('pol_embedding_size', self.RESIDUAL_FILTERS)
+        self.val_embedding_size = self.cfg['model'].get('value_embedding_size', 32)
+        self.mov_embedding_size = self.cfg['model'].get('moves_left_embedding_size', 8)
+        #policy head
         self.pol_encoder_layers = self.cfg['model'].get('pol_encoder_layers', 1)
-        self.pol_encoder_heads = self.cfg['model'].get('pol_encoder_heads', 2)
+        self.pol_encoder_heads = self.encoder_heads
```
Tilps, line 161: I think this should be self.cfg['model'].get('pol_encoder_heads', self.encoder_heads) ?

tf/net.py
```diff
@@ -51,15 +52,19 @@ def __init__(self,
...
     def set_networkformat(self, net):
         self.pb.format.network_format.network = net
+        if net == pb.NetworkFormat.NETWORK_ATTENTIONBODY_WITH_HEADFORMAT \
+                and self.pb.min_version.minor < LC0_MINOR_WITH_ATTN_BODY:
+            self.pb.min_version.minor = LC0_MINOR_WITH_ATTN_BODY
 
     def set_policyformat(self, policy):
         self.pb.format.network_format.policy = policy
 
-    def set_pol_headcount(self, headcount):
```
Tilps, line 61: I think you want both for backwards compatibility reasons.

tf/net.py
```diff
@@ -348,7 +407,7 @@ def moves_left_to_bp(l, w):
         weights_name = layers[-1]
         pb_name = None
         block = None
-        pol_encoder_block = None
+        encoder_block = None
```
Tilps, line 410: I think you want both.

tf/net.py
```diff
@@ -365,7 +424,7 @@ def moves_left_to_bp(l, w):
             elif layers[1] == 'attention':
                 pb_name = attn_pol_to_bp(layers[2], weights_name)
             elif layers[1].startswith('enc_layer_'):
-                pol_encoder_block = int(layers[1].split('_')[2]) - 1
+                encoder_block = int(layers[1].split('_')[2]) - 1
```
Tilps, line 427: Don't change this...

tf/net.py
```diff
@@ -392,8 +451,32 @@ def moves_left_to_bp(l, w):
...
+            elif layers[1].split(':')[0] == 'bias':
+                pb_name = 'ip_emb_b'
+        elif base_layer == 'smol_weight_gen':
+            if layers[1].split(':')[0] == 'kernel':
+                pb_name = 'smolgen_w'
+            else:
+                pb_name = 'smolgen_b'
 
-        return (pb_name, block, pol_encoder_block)
+        return (pb_name, block, encoder_block)
```
Tilps, line 479: Pass both.

tf/net.py
```diff
@@ -412,18 +495,18 @@ def get_weights_v2(self, names):
...
+            pb_name, block, encoder_block = self.tf_name_to_pb_name(name)
 
             if pb_name is None:
                 raise ValueError(
                     "Don't know where to store weight in protobuf: {}".format(
                         name))
 
             if block is None:
-                if pol_encoder_block is None:
+                if encoder_block is None:
```
Tilps, line 506: Handle both.

tf/net.py
```diff
@@ -555,25 +638,28 @@ def fill_net_v2(self, all_weights):
...
```
Tilps, line 656: Handle both.

tf/tfprocess.py
```diff
@@ -1135,61 +1197,122 @@ def residual_block(self, inputs, channels, name):
...
         # split q, k and v into smaller vectors of size 'depth' -- one for each head in multi-head attention
         batch_size = tf.shape(q)[0]
+
         q = self.split_heads(q, batch_size, num_heads, depth)
         k = self.split_heads(k, batch_size, num_heads, depth)
         v = self.split_heads(v, batch_size, num_heads, depth)
-        scaled_attention, attention_weights = self.scaled_dot_product_attention(q, k, v)
-        if num_heads > 1:
-            scaled_attention = tf.transpose(scaled_attention, perm=[0, 2, 1, 3])
-            scaled_attention = tf.reshape(scaled_attention, (batch_size, -1, d_model))  # concatenate heads
```
Tilps, line 1168: Curious this reshape is not the same as the one now embedded into the called function...

tf/tfprocess.py
```diff
@@ -1135,61 +1197,122 @@ def residual_block(self, inputs, channels, name):
...
+
+        scaled_attention, attention_weights = self.scaled_dot_product_attention(
+            q, k, v, name=name, inputs=inputs)
+
         # final dense layer
-        output = tf.keras.layers.Dense(emb_size, kernel_initializer='glorot_normal',
-                                       name=name + "/dense")(scaled_attention)
+        output = tf.keras.layers.Dense(
+            emb_size, name=name + "/dense_fin", kernel_initializer=initializer)(scaled_attention)
```
Tilps, line 1255: Renamed weights - but no obvious corresponding change in the network save logic?

tf/tfprocess.py
```diff
@@ -1135,61 +1197,122 @@ def residual_block(self, inputs, channels, name):
...
     # 2-layer dense feed-forward network in encoder blocks
-    def ffn(self, inputs, emb_size, dff, name):
-        dense1 = tf.keras.layers.Dense(dff, kernel_initializer='glorot_normal', activation='selu',
-                                       name=name + "/dense1")(inputs)
-        return tf.keras.layers.Dense(emb_size, kernel_initializer='glorot_normal', name=name + "/dense2")(dense1)
+    def ffn(self, inputs, emb_size: int, dff: int, initializer, name: str):
+        dense1 = tf.keras.layers.Dense(
+            dff, name=name + "/dense1", kernel_initializer=initializer)(inputs)
+
+        activation = square_relu if self.square_relu_ffn else tf.keras.activations.get(self.DEFAULT_ACTIVATION)
```
Tilps, line 1263: Seems curious not to pass this fused to the dense above?

tf/tfprocess.py
```diff
@@ -1135,61 +1197,122 @@ def residual_block(self, inputs, channels, name):
...
+        dense1 = activation(dense1)
+
+        out = tf.keras.layers.Dense(emb_size, name=name + "/dense2", kernel_initializer=initializer)(dense1)
+        return out
+
+    def encoder_layer(self, inputs, emb_size: int, d_model: int, num_heads: int, dff: int, name: str, training: bool):
+        # DeepNorm
+        alpha = tf.cast(tf.math.pow(
```
Tilps, line 1271: Don't we need a path where alpha is forced to 1 for policy head encoder backwards compat for the non-attention body case?

tf/tfprocess.py
masterkni6: reverted back, not sure if there is a difference in the end

tf/tfprocess.py
```diff
@@ -1199,6 +1324,80 @@ def construct_net(self, inputs):
...
+                                     name='embedding')(flow)
+
+        # !!! input gate
+        flow = ma_gating(flow, name='embedding')
+
+        for i in range(self.encoder_layers):
+            flow, attn_wts_l = self.encoder_layer(flow, embedding_size, self.encoder_d_model,
+                                                  self.encoder_heads, self.encoder_dff,
+                                                  name='encoder_{}'.format(i + 1))
+        return flow, attn_wts_l
```
almaudoh, line 1360: This `attn_wts_l` for each encoder block should be added into an array and returned. Like in line 1449 below.

masterkni6: added them all to the same list. Any attention visualizers will have to account for that :P

tf/net.py
```diff
@@ -51,13 +52,20 @@ def __init__(self,
         self.set_valueformat(value)
         self.set_movesleftformat(moves_left)
         self.set_defaultactivation(pb.NetworkFormat.DEFAULT_ACTIVATION_RELU)
+        # self.set_smolgen_activation(pb.Weights.Activation.SWISH)
```
Tilps, line 55: Delete?

tf/net.py
```diff
@@ -412,18 +499,24 @@ def get_weights_v2(self, names):
...
 
             if pb_name is None:
                 raise ValueError(
                     "Don't know where to store weight in protobuf: {}".format(
                         name))
 
             if block is None:
-                if pol_encoder_block is None:
-                    pb_weights = self.pb.weights
+                if name.startswith("policy"):
```
Tilps, line 510: I think a given weight can only have at most one of pol_encoder_block or encoder_block set to not None, so you can probably just nest the checks rather than needing to check the name.

tf/net.py
```diff
@@ -555,25 +648,37 @@ def fill_net_v2(self, all_weights):
...
```
Tilps, line 666: Same as above.

tf/tfprocess.py
```diff
@@ -32,6 +32,35 @@
...
+
+    def call(self, inputs):
+        return tf.add(inputs, self.gate) if self.additive else tf.multiply(inputs, self.gate)
+
+
+def ma_gating(inputs, name):
+    out = Gating(name=name+'/mult_gate', additive=False)(inputs)
+    out = Gating(name=name+'/add_gate', additive=True)(out)
+    return out
+
```
Tilps, line 63: I think maybe the file needs to run through yapf?

tf/tfprocess.py
```diff
@@ -157,6 +211,8 @@ def __init__(self, cfg):
             self.POLICY_HEAD = pb.NetworkFormat.POLICY_CONVOLUTION
         elif policy_head == "attention":
             self.POLICY_HEAD = pb.NetworkFormat.POLICY_ATTENTION
+            if self.encoder_layers > 0:
```
Tilps, line 214: Can this be integrated into the similar if check on line 279 instead?

tf/tfprocess.py
```diff
@@ -220,6 +276,14 @@ def __init__(self, cfg):
             raise ValueError(
                 "Unknown default activation type: {}".format(default_activation))
 
+        if self.encoder_layers > 0:
+            self.net.set_networkformat(
+                pb.NetworkFormat.NETWORK_ATTENTIONBODY_WITH_HEADFORMAT)
+
+        self.net.set_smolgen_activation(self.net.activation(self.smolgen_activation))
```
Tilps, line 283: These two sets up the minimum version of the network - you should guard them by the network format being attention body with head format.

tf/tfprocess.py
```diff
@@ -1199,6 +1324,81 @@ def construct_net(self, inputs):
             flow = self.residual_block(flow,
                                        self.RESIDUAL_FILTERS,
                                        name='residual_{}'.format(i + 1))
+        return flow
+
+    def create_encoder_body(self, inputs, embedding_size):
+        # Policy head
+        assert self.POLICY_HEAD == pb.NetworkFormat.POLICY_ATTENTION
+
+        # TODO: re-add support for policy encoder blocks
```
Tilps, line 1333: TODO can be removed?

tf/tfprocess.py
```diff
@@ -1229,69 +1429,49 @@ def construct_net(self, inputs):
...
-
-            # ENCODER LAYERS: intermediate layers of self-attention with residual connections
-            attn_wts = []
-            for i in range(self.pol_encoder_layers):
-                tokens, attn_wts_l = self.encoder_layer(tokens, self.pol_embedding_size, self.pol_encoder_d_model,
-                                                        self.pol_encoder_heads, self.pol_encoder_dff,
-                                                        name='policy/enc_layer_{}'.format(i + 1), training=True
-                                                        )
-                attn_wts.append(attn_wts_l)
+            if self.RESIDUAL_BLOCKS > 0:
```
Tilps, line 1444: This implies we need an assert somewhere that we don't actually support the combination of encoders and policy encoders?
We might want an assert that residual blocks is 0 if encoders is non-zero and vice versa as well? (If I didn't miss it)

masterkni6: line 149 should prevent a combination of body encoder and residual blocks
```
assert (self.RESIDUAL_BLOCKS > 0) != (self.encoder_layers > 0), \
                "Nets with both encoder layers and residual blocks are not supported"
```

tf/tfprocess.py
masterkni6: not sure what is wanted here, so i put the two activations into the above if statement

tf/tfprocess.py
```diff
@@ -106,23 +145,62 @@ def reset(self):
...
-        self.pol_encoder_layers = self.cfg['model'].get('pol_encoder_layers', 1)
+        self.pol_embedding_size = self.cfg['model'].get(
+            'pol_embedding_size', self.RESIDUAL_FILTERS)
+        self.val_embedding_size = self.cfg['model'].get(
+            'value_embedding_size', 32)
+        self.mov_embedding_size = self.cfg['model'].get(
+            'moves_left_embedding_size', 8)
+        #policy head
+        self.pol_encoder_layers = self.cfg['model'].get(
+            'pol_encoder_layers', 1)
```
Tilps, line 176: Lets make this default to 0 if encoder_layers > 0.
Then add assert encoder_layers and pol_encoder_layers are not both > 0

Tilps (approved):

tf/tfprocess.py
masterkni6: added with explict asserts for users that add both body encoder and pol encoder in yaml

masterkni6:
net_to_model.py and then model_to_net.py yields a working t81 net on v0.29.0

PR #216, opened 2023-04-16
update use of aggregator logic

masterkni6:
Update logic to account for legacy opt requiring gradient_aggregator

### DIFF
--- a/tf/tfprocess.py
+++ b/tf/tfprocess.py
@@ -400,6 +400,10 @@ class TFProcess:
             self.optimizer = tf.keras.optimizers.SGD(
                 learning_rate=self.active_lr, momentum=0.9, nesterov=True)
             self.update_lr_manually = True
+            try:
+                self.aggregator = self.optimizer.aggregate_gradients
+            except AttributeError:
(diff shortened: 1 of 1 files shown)


tf/tfprocess.py
```diff
@@ -400,6 +400,10 @@ def init_net(self):
             self.optimizer = tf.keras.optimizers.SGD(
                 learning_rate=self.active_lr, momentum=0.9, nesterov=True)
             self.update_lr_manually = True
+            try:
```
masterkni6, line 403: technically this should only be 
`self.aggregator = self.optimizer.aggregate_gradients`
because new optimizer only exists in later tf

tf/tfprocess.py
```diff
@@ -400,6 +400,10 @@ def init_net(self):
...
+                self.aggregator = self.optimizer.aggregate_gradients
+            except AttributeError:
+                self.aggregator = self.optimizer.gradient_aggregator
```
masterkni6, line 406: probably try block is not needed since this is for new optimizers

PR #239, opened 2026-03-17
Improve data loading multi thread performance

john-sp:
This PR combines two improvements to the data loading pipeline:
1. Moves mutex protection to the file source level in the tar chunk reader to reduce contention and improve thread safety when reading from multiple files concurrently.
2. Adds batch prefetching in the Python training loop to overlap data preparation with computation, improving training throughput. (tested only a few seconds faster per 5k steps)
Additionally, the C++ chunk source loader was modernized to use file descriptors with pread() for thread-safe reading, replaced FILE* usage, added proper mutex protection for shared data structures, and used shared_ptr for better lifetime management across threads. These changes collectively improve the performance and reliability of the data loading system.

### DIFF
--- a/csrc/loader/chunk_source/tar_chunk_source.cc
+++ b/csrc/loader/chunk_source/tar_chunk_source.cc
@@ -7,11 +7,9 @@
 #include <algorithm>
 #include <array>
 #include <cassert>
-#include <cerrno>
 #include <cstdint>
 #include <cstring>
 #include <fcntl.h>
-#include <limits>
 #include <stdexcept>
 #include <unistd.h>
 
@@ -52,24 +50,12 @@ uint64_t ParseOctal(const std::array<uint8_t, 12>& octal) {
 }
 
 bool ReadExact(int fd, off_t offset, void* buffer, size_t size) {
-  if (offset < 0) return false;
-  if (size > static_cast<size_t>(std::numeric_limits<ssize_t>::max())) {
-    return false;
-  }
   char* out = static_cast<char*>(buffer);
   size_t read_total = 0;
   while (read_total < size) {
-    if (offset > std::numeric_limits<off_t>::max() -
-                     static_cast<off_t>(read_total)) {
-      return false;
-    }
     const ssize_t read_now =
         pread(fd, out + read_total, size - read_total, offset + read_total);
-    if (read_now < 0) {
-      if (errno == EINTR) continue;
-      return false;
-    }
-    if (read_now == 0) return false;
+    if (read_now <= 0) return false;
     read_total += static_cast<size_t>(read_now);
   }
   return true;
@@ -154,12 +140,10 @@ TarChunkSource::TarChunkSource(
 TarChunkSource::~TarChunkSource() { Close(); }
 
 void TarChunkSource::Close() {
-  if (fd_ >= 0) {
-    if (close(fd_) != 0) {
-      PLOG(WARNING) << "Failed to close tar file descriptor for " << path_;
-    }
-    fd_ = -1;
+  if (fd_ >= 0 && close(fd_) != 0) {
+    PLOG(WARNING) << "Failed to close tar file descriptor for " << path_;
   }
+  fd_ = -1;
 }
 
 std::string TarChunkSource::GetChunkSortKey() const { return filename_; }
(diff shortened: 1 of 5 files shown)


csrc/loader/chunk_source/tar_chunk_source.cc
```diff
@@ -46,13 +51,33 @@ uint64_t ParseOctal(const std::array<uint8_t, 12>& octal) {
   return value;
 }
 
-std::optional<std::string> ReadGzipPrefix(FILE* file, long int offset,
-                                          long int size, size_t max_bytes) {
-  if (max_bytes == 0) return std::string();
-
-  if (fseek(file, offset, SEEK_SET) != 0) {
-    return std::nullopt;
+bool ReadExact(int fd, off_t offset, void* buffer, size_t size) {
```
mooskagh, line 52: This function takes offset but doesn't use it?

csrc/loader/chunk_source/tar_chunk_source.cc
```diff
@@ -46,13 +51,33 @@ uint64_t ParseOctal(const std::array<uint8_t, 12>& octal) {
...
+  if (offset < 0) return false;
+  if (size > static_cast<size_t>(std::numeric_limits<ssize_t>::max())) {
```
mooskagh: Is this check needed for something? Looks quite useless.

csrc/loader/chunk_source/tar_chunk_source.cc
mooskagh: Ah sorry, I accidentally included comma in the search.

csrc/loader/chunk_source/tar_chunk_source.cc
```diff
@@ -46,13 +51,33 @@ uint64_t ParseOctal(const std::array<uint8_t, 12>& octal) {
...
+    return false;
   }
+  char* out = static_cast<char*>(buffer);
+  size_t read_total = 0;
+  while (read_total < size) {
+    if (offset > std::numeric_limits<off_t>::max() -
```
mooskagh: This also appears quite useless..

csrc/loader/chunk_source/tar_chunk_source.cc
```diff
@@ -46,13 +51,33 @@ uint64_t ParseOctal(const std::array<uint8_t, 12>& octal) {
...
+                     static_cast<off_t>(read_total)) {
+      return false;
+    }
+    const ssize_t read_now =
+        pread(fd, out + read_total, size - read_total, offset + read_total);
+    if (read_now < 0) {
+      if (errno == EINTR) continue;
```
mooskagh: not needed. And then we can collapse condition to `read_now <= 0`

csrc/loader/chunk_source/tar_chunk_source.cc
```diff
@@ -46,13 +51,33 @@ uint64_t ParseOctal(const std::array<uint8_t, 12>& octal) {
...
 }
 
-std::optional<std::string> ReadGzipPrefix(FILE* file, long int offset,
-                                          long int size, size_t max_bytes) {
-  if (max_bytes == 0) return std::string();
-
-  if (fseek(file, offset, SEEK_SET) != 0) {
-    return std::nullopt;
+bool ReadExact(int fd, off_t offset, void* buffer, size_t size) {
+  if (offset < 0) return false;
```
mooskagh: That's also not needed. We can at most use assert()..

csrc/loader/chunk_source/tar_chunk_source.cc
```diff
@@ -113,29 +138,44 @@ std::optional<std::string> ReadGzipPrefix(FILE* file, long int offset,
...
   Index();
 }
 
-TarChunkSource::~TarChunkSource() {
-  if (file_) fclose(file_);
+TarChunkSource::~TarChunkSource() { Close(); }
+
+void TarChunkSource::Close() {
+  if (fd_ >= 0) {
+    if (close(fd_) != 0) {
```
mooskagh: I'd combine two `if`s into one. (and assign -1 unconditionally)

csrc/loader/chunk_source/tar_chunk_source.cc
```diff
@@ -151,19 +191,23 @@ void TarChunkSource::Index() {
...
-    const long int offset = ftell(file_);
-    const long int size = ParseOctal(header.size);
-    const long int new_offset = offset + (size + 511) / 512 * 512;
-    fseek(file_, new_offset, SEEK_SET);
-    if (new_offset != ftell(file_)) {
+    const off_t file_offset = offset;
+    const size_t size = ParseOctal(header.size);
+    const uint64_t padded_size =
+        ((static_cast<uint64_t>(size) + 511) / 512) * 512;
+    if (padded_size > static_cast<uint64_t>(std::numeric_limits<off_t>::max()) ||
```
mooskagh: These check for overflows are not needed either. (also withoug ftell we won't know it's truncated before we read, so this LOG(WARNING) has to be removed too)

src/lczero_training/training/training.py
mooskagh, line 1: What is the intention here? It doesn't appear the changes in this file are useful enough to justify adding complexity?

mooskagh (changes requested):

src/lczero_training/training/training.py
john-sp: This is loading the next batch to the GPU while the current batch is running, saved me roughly 30 seconds per 5k steps (40 minutes).

mooskagh (approved):
csrc/loader/stages/shuffling_chunk_pool.cc is a mess of course (even before this change)... I hope to clean it up one day.

src/lczero_training/training/training.py
mooskagh: But we don't have 30 seconds per step? Our training step is like 1/3 of a second?

mooskagh (changes requested):

src/lczero_training/training/training.py
john-sp: It was saving 0.006 seconds per step, that is most like within the training noise. Reverted this file

mooskagh (approved):

PR #245, opened 2026-07-11
Update Jax Version

john-sp:
Update JAX version to latest.

Will update description when fully tested on Multi-GPU system for training.


pyproject.toml
```diff
@@ -13,14 +13,14 @@ dependencies = [
     "mypy>=1.17.1",
     "pytest>=8.4.1",
     "anyio>=4.10.0",
-    "jax[cuda12]==0.9.1",
+    "jax[cuda12]==0.10.2",
```
mooskagh, line 16: Note that jax also has `[cuda13]` variant, which may be better for newer GPUs (for A100 and earlier cuda12 is probably still better, and pre-Ampere is dropped from cuda 13 completely)

mooskagh (approved):

pyproject.toml
mooskagh: Ah and also there's jax 0.11 now I believe.

